PR #6, opened 2020-12-10
add command for list exps

WangXiangUSTC:

users may forget to save the uid after doing attack, need to list all the experiments


WangXiangUSTC:
@cwen0 PTAL

cwen0:
@WangXiangUSTC We have implemented a search command to list the experiment.

WangXiangUSTC:
> @WangXiangUSTC We have implemented a search command to list the experiment. If we want to list all experiments, we can run the following command:
> 
> ```
> chaosd search -A
> ```

./bin/chaosd search -A
UID is required

PR #9, opened 2020-12-24
support create chaos and recover chaos in server mode

WangXiangUSTC:
example:

create attack:
```bash
curl -X POST 127.0.0.1:31767/api/attack/stress -H "Content-Type: application/json" -d '{"action":"cpu"}'
```

recover attack:
```bash
curl -X DELETE 127.0.0.1:31767/api/attack/bcfefb35-aafe-4e16-904e-20a1140f2b4c
```


WangXiangUSTC:
@cwen0 PTAL

pkg/server/utils/recover.go
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"context"
+	"encoding/json"
+
+	//"fmt"
```
cwen0, line 20: format this

pkg/server/utils/recover.go
```diff
@@ -0,0 +1,80 @@
...
+	"syscall"
+
+	"github.com/pkg/errors"
+
+	"github.com/chaos-mesh/chaosd/pkg/core"
+	"github.com/chaos-mesh/chaosd/pkg/server/chaosd"
+)
+
+func RecoverExp(expStore core.ExperimentStore, chaos *chaosd.Server, uid string) error {
```
cwen0, line 29: Why define this function in `utils` pkg?

WangXiangUSTC: `RecoverExp` is used both in `server` mode and  `command` mode, so put it in `utils`

cwen0 (approved):
LGTM

PR #12, opened 2021-01-05
Add tests in utils

ClSlaid:
add a few tests in chaosd/pkg/utils
also fixed typo in path.go
well maybe this is way too simple

### DIFF
--- a/pkg/utils/cidr.go
+++ b/pkg/utils/cidr.go
@@ -20,8 +20,15 @@ import (
 
 // IPToCidr converts from an ip to a full mask cidr
 func IPToCidr(ip string) string {
-	// TODO: support IPv6
-	return ip + "/32"
+	// TODO: support IPv6 correctly
+
+	// distinguish is IPv4 or IPv6 address
+	// no error checking here!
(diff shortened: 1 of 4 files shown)


pkg/utils/cidr.go
```diff
@@ -20,8 +20,14 @@ import (
 
 // IPToCidr converts from an ip to a full mask cidr
 func IPToCidr(ip string) string {
-	// TODO: support IPv6
-	return ip + "/32"
+	// TODO: support IPv6 correctly
```
WangXiangUSTC, line 23: do you mean the support of IPV6 is not correctly now? 😃

ClSlaid: I don't really consider merely appending a "/32" or "/128" to an IP address a correct to convert IP to CIDR. As I know a CIDR more likely refer to a *Range* while IP is to a *Dot*.

WangXiangUSTC: ok, we can fix it later

WangXiangUSTC (approved):
thanks, LGTM

PR #17, opened 2021-02-03
 set config for sqlite to avoid error `database is locked`

WangXiangUSTC:
same with https://github.com/chaos-mesh/chaos-mesh/pull/1411


cwen0 (approved):
LGTM

PR #23, opened 2021-03-03
Add test for process attack

### LINKED ISSUE #19: add some test for process chaos
We need to add some tests for process chaos, like integration test or unit test, can refer to https://github.com/chaos-mesh/chaosd/tree/main/test/integration_test

AsterNighT:
resolve #19 
Can definitely use some help on file structure.

### DIFF
--- a/test/integration_test/process/run.sh
+++ b/test/integration_test/process/run.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 Chaos Mesh Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -u
+
+cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+cd $cur
+
+bin_path=../../../bin
+
+${bin_path}/dummy > dummy.out &
+
+sleep 1
+
+pid=$(cat dummy.out)
+
+echo ${pid}
+
+# test stop attack
+${bin_path}/chaosd attack process stop -p ${pid} > proc.out
+
+sleep 1
+
+uid=$(cat proc.out | grep "Attack process ${pid} successfully" | awk -F: '{print $2}')
+
+stat=$(ps o pid,s | grep ${pid} | awk '{print $2}')
+
+if [[ ${stat} != "T" ]]; then
+    echo "target process is not stopped by processed stop attack"
+    exit 1
+fi
+
+# test recover from stop attack
+${bin_path}/chaosd recover ${uid}
+
+sleep 1
+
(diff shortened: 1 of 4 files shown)


test/integration_test/process/run.sh
```diff
@@ -0,0 +1,72 @@
...
+# test stop attack
+${bin_path}/chaosd attack process stop -p ${pid} > proc.out
+
+sleep 1
+
+uid=$(cat proc.out | grep "Attack process ${pid} successfully" | awk -F: '{print $2}')
+
+stat=$(ps axo pid,s | grep ${pid} | awk '{print $2}')
+
+if [[ ${stat} != "T" && ${stat} != "t" ]]; then
```
WangXiangUSTC, line 38: how about adding a comment of what the `t` means?

WangXiangUSTC (commented):
rest LGTM

WangXiangUSTC:
@cwen0 PTAL

test/utilities/dummy.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"os"
+)
+
+func main() {
+	fmt.Println(os.Getpid())
+	for {
+		// Need to bypass revive
+		a := 1
+		a = a + a
```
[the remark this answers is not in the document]
AsterNighT: Indeed, sleep would be better

test/integration_test/process/run.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+# test recover from stop attack
+${bin_path}/chaosd recover ${uid}
+
+sleep 1
+
+stat=$(ps axo pid,s | grep ${pid} | awk '{print $2}')
+
+if [[ ${stat} != "R" ]]; then
```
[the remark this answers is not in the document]
AsterNighT: Fixed

test/integration_test/process/run.sh
AsterNighT: Fixed

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
https://github.com/Andrewmatilde PTAL

test/integration_test/process/run.sh
```diff
@@ -0,0 +1,74 @@
...
+sleep 1
+
+stat=$(ps axo pid | grep ${pid})
+
+if [[ -n ${stat} ]]; then
+    echo "target process is not killed by processed kill attack"
+    exit 1
+fi
+
+rm dummy.out
```
Andrewmatilde, line 69: I suggested give way to clear & direct delete dummy.out & proc.out when error exit.

Andrewmatilde (commented):
Thanks for your pull request.

test/integration_test/process/run.sh
AsterNighT: Umm, wouldn't that be useful for debugging if the test does fail?

Andrewmatilde: But it just leave an empty file ( dummy.out ) and a file with just a pid ( proc.out ) without any log info when test does fail. I think it's hard to say it works.

Andrewmatilde: It may because that attack process lacks log info.

AsterNighT: hmm you got me on that

Andrewmatilde:
LGTM

PR #26, opened 2021-03-09
feat: support shutdown

### LINKED ISSUE #24: support shutdown
chaosd needs to support shutdown the host machine which the chaosd executed on.

just execute the `shutdown` command in chaosd is OK.

shivanshs9:
Fixes #24


pkg/server/chaosd/host_attack.go
```diff
@@ -0,0 +1,62 @@
...
+		Status:         core.Created,
+		Kind:           core.HostAttack,
+		Action:         attack.Action,
+		RecoverCommand: attack.String(),
+	}); err != nil {
+		err = errors.WithStack(err)
+		return
+	}
+
+	defer func() {
```
shivanshs9, line 45: @WangXiangUSTC This function that's deferred could be made a utility function since it's used a lot?

WangXiangUSTC: Yes, I agree with you

pkg/server/chaosd/host_attack.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2020 Chaos Mesh Authors.
```
WangXiangUSTC, line 1:
```suggestion
// Copyright 2021 Chaos Mesh Authors.
```

pkg/server/chaosd/host_attack_unix.go
```diff
@@ -0,0 +1,35 @@
+// +build aix darwin dragonfly freebsd js,wasm linux nacl netbsd openbsd solaris
+
+// Copyright 2020 Chaos Mesh Authors.
```
WangXiangUSTC, line 3:
```suggestion
// Copyright 2021 Chaos Mesh Authors.
```

pkg/server/chaosd/host_attack.go
shivanshs9: So should I include all these refactor changes here or in another PR?

WangXiangUSTC: I think it's better to refactor it in another PR

pkg/server/chaosd/host_attack_unix.go
```diff
@@ -0,0 +1,35 @@
...
+
+const CmdShutdown = "shutdown"
+
+func (h UnixHost) Name() string {
+	return "unix"
+}
+
+func (h UnixHost) Shutdown() error {
+	cmd := exec.Command(CmdShutdown)
+	return cmd.Run()
```
WangXiangUSTC, line 34:
```suggestion
        output, err := cmd.CombinedOutput()
        if err != nil {
            log.Error(string(output), zap.Error(err))
            return err
        }
        return nil
	
```

WangXiangUSTC: how about using `CombinedOutput`? so we can get the reason why it executes failed

WangXiangUSTC (commented):
rest LGTM

pkg/server/chaosd/host_attack_unix.go
shivanshs9: Fixed :+1:

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
@Andrewmatilde PTAL

Andrewmatilde:
LGTM

Issue #28, opened 2021-03-10
support IO chaos on disk


WangXiangUSTC:
Note: It is different from the IO chaos in Chaos Mesh

This feature includes the following capabilities：
- emulation of real read and write on disk
- fill the disk

WangXiangUSTC:
@Andrewmatilde PTAL

Andrewmatilde:
OK,I agree with you.

YangKeao:
There is another problem. @Andrewmatilde  @WangXiangUSTC Do you think Chaos Mesh should also provide similar functions? Should we brought this feature to `chaos-daemon`?

WangXiangUSTC:
> There is another problem. @Andrewmatilde @WangXiangUSTC Do you think Chaos Mesh should also provide similar functions? Should we brought this feature to `chaos-daemon`?

I think we can do this

PR #30, opened 2021-03-16
Support disk fill & add disk (read|write) payload

Andrewmatilde:
Related: #28 
### Requirement:
Only support unix-like system.
### USAGE:
```
$ ./bin/chaosd attack disk -h
disk attack related command

Usage:
  chaosd attack disk [command]

Available Commands:
  add-payload add disk payload
  fill        fill disk

Flags:
  -h, --help   help for disk

Global Flags:

Use "chaosd attack disk [command] --help" for more information about a **command.
```
### Fill disk
```
$ ./bin/chaosd attack disk fill -h
fill disk

Usage:

Flags:
  -h, --help          help for fill
  -p, --path string   'path' specifies the location to fill data in.
                      If path not provided, a temp file will be generated and delete after data filled in
  -s, --size uint     'size' specifies how many data will fill in the file path with unit MB.

Global Flags:
```
### add read payload
```
$ ./bin/chaosd attack disk add-payload read -h
read payload

Usage:
  chaosd attack disk add-payload read [flags]

Flags:
  -h, --help          help for read
  -p, --path string   'path' specifies the location to read data.
                      If path not provided, payload will read from /dev/sda (a disk device) (default "/dev/sda")
  -s, --size uint     'size' specifies how many data will read from the file path with unit MB.

Global Flags:
```
### add write payload
```
$ ./bin/chaosd attack disk add-payload write -h
write payload

Usage:
  chaosd attack disk add-payload write [flags]

Flags:
  -h, --help          help for write
  -p, --path string   'path' specifies the location to fill data in.
                      If path not provided, payload will write into /dev/null (default "/dev/null")
  -s, --size uint     'size' specifies how many data will fill in the file path with unit MB.

Global Flags:
```


Andrewmatilde:
@WangXiangUSTC  PTAL

pkg/server/chaosd/disk.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		return uid, err
+	default:
+		err := errors.Errorf("invalid fill action")
+		log.Error(fmt.Sprintf("fill action: %s", fill.Action), zap.Error(err))
+		return uid, err
+	}
+}
+
+const DDFillCommand = "dd if=/dev/zero of=%s bs=%s count=%s iflag=fullblock"
```
YangKeao, line 94: `dd` will take a lot of time to fill up a big volume. How about use "syscall.Fallocate()" to allocate a big file immediately? Both of `ext4` and `xfs` support this syscall.

Andrewmatilde: Thanks for your review, I agree with you. I will just add a fallocate option later & set it to default method, because fallocate not work in some place.

WangXiangUSTC:
how about adding some test for it? Either Integration test or Unit test is OK

Andrewmatilde:
> how about adding some test for it? Either Integration test or Unit test is OK

OK. I will add some Unit test

cmd/chaosd/ctl/command/dick_command.go
```diff
@@ -0,0 +1,142 @@
...
+func NewDiskReadPayloadCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "read",
+		Short: "read payload",
+		Run:   ReadDiskPayloadCommandFunc,
+	}
+
+	cmd.Flags().Uint64VarP(&dFlag.Size, "size", "s", 0,
+		"'size' specifies how many data will read from the file path with unit MB.")
+	cmd.Flags().StringVarP(&dFlag.Path, "path", "p", "/dev/sda",
```
WangXiangUSTC, line 91: Why the default value is `/dev/sda`?

cmd/chaosd/ctl/command/dick_command.go
```diff
@@ -0,0 +1,142 @@
...
+		ExitWithError(ExitBadArgs, err)
+	}
+	chaos := mustChaosdFromCmd(cmd, &conf)
+
+	uid, err := chaos.DiskPayload(&dFlag)
+	if err != nil {
+		ExitWithError(ExitError, err)
+	}
+
+	NormalExit(fmt.Sprintf("Fill file %s successfully, uid: %s", dFlag.Path, uid))
```
WangXiangUSTC, line 109:
```suggestion
	NormalExit(fmt.Sprintf("Read file %s successfully, uid: %s", dFlag.Path, uid))
```

pkg/server/chaosd/disk.go
```diff
@@ -0,0 +1,169 @@
...
+			log.Error("failed to update experiment", zap.Error(err))
+		}
+	}()
+
+	switch fill.Action {
+	case core.DiskWritePayloadAction:
+		if fill.Path == "" {
+			fill.Path = "/dev/null"
+		}
+		cmd := exec.Command("bash", "-c", fmt.Sprintf(DDWritePayloadCommand, fill.Path, "1M", strconv.FormatUint(fill.Size, 10)))
```
WangXiangUSTC, line 65: What is the `1M` means?

cmd/chaosd/ctl/command/dick_command.go
```diff
@@ -0,0 +1,142 @@
+// Copyright 2020 Chaos Mesh Authors.
```
WangXiangUSTC, line 1:
```suggestion
// Copyright 2021 Chaos Mesh Authors.
```

pkg/core/disk.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2020 Chaos Mesh Authors.
```
WangXiangUSTC, line 1:
```suggestion
// Copyright 2021 Chaos Mesh Authors.
```

pkg/server/chaosd/disk.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2020 Chaos Mesh Authors.
```
WangXiangUSTC, line 1:
```suggestion
// Copyright 2021 Chaos Mesh Authors.
```

cmd/chaosd/ctl/command/dick_command.go
Andrewmatilde: /dev/sda is the default disk device name in Ubuntu. I may delete this default value because this value must be set by user.

pkg/server/chaosd/disk.go
Andrewmatilde: 1M means the block size. The bytes size dd read | write is block size * size.

WangXiangUSTC: OK, how about adding a  comment for it?

Andrewmatilde: OK,I agree with you.

pkg/server/chaosd/disk.go
```diff
@@ -0,0 +1,172 @@
...
+			return
+		}
+		if err := s.exp.Update(context.Background(), uid, core.Success, "", fill.String()); err != nil {
+			log.Error("failed to update experiment", zap.Error(err))
+		}
+	}()
+
+	if fill.Path == "" {
+		for {
+			tempFilename := rand.String(128) + "temp-data"
```
WangXiangUSTC, line 126: how about using [TempFile](https://golang.org/pkg/io/ioutil/#TempFile)?

WangXiangUSTC (commented):
rest LGTM

pkg/server/chaosd/disk.go
Andrewmatilde: OK, I agree with you.

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
@YangKeao PTAL again

pkg/server/chaosd/disk.go
```diff
@@ -0,0 +1,164 @@
...
+	}()
+
+	if fill.Path == "" {
+		tempFile, err := ioutil.TempFile("", "example")
+		if err != nil {
+			log.Error("unexpected err when open temp file", zap.Error(err))
+			return uid, err
+		}
+
+		if tempFile != nil {
```
WangXiangUSTC, line 131: the `tempFile` shouldn't be nil if err is nil in L125, otherwise, we need to move L138 - L144 to this block.

Andrewmatilde: OK, I agree with you . I will raise an error when tempfile get nil file return .

WangXiangUSTC (approved):
LGTM

cmd/chaosd/ctl/command/dick_command.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func NewDiskFillCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "fill",
+		Short: "fill disk",
+		Run:   DiskFillCommandFunc,
+	}
+
+	dFlag.FillByFallocate = true
```
YangKeao, line 119: The default value has already been set in `cmd.Flags().BoolVarP`, doesn't it?

Andrewmatilde: Sorry , I miss it.

YangKeao (approved):
LGTM

PR #32, opened 2021-03-17
fix unstable stress test

WangXiangUSTC:

### DIFF
--- a/test/integration_test/stress/run.sh
+++ b/test/integration_test/stress/run.sh
@@ -23,6 +23,8 @@ bin_path=../../../bin
 # test cpu stress
 ${bin_path}/chaosd attack stress cpu -l 10 -w 1 > cpu.out
 
+PID=`cat cpu.out | grep "stress-ng" | sed 's/.*Pid=\([0-9]*\).*/\1/g'`
+
 stress_ng_num=`ps aux > test.temp && grep "stress-ng" test.temp | wc -l && rm test.temp`
 if [ ${stress_ng_num} -lt 1 ]; then
     echo "stress-ng is not run when executing stress cpu attack"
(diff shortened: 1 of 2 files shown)


WangXiangUSTC:
@Andrewmatilde PTAL

Andrewmatilde:
I am confused about how this problem works. Do we need new a issue or add some message about it?

WangXiangUSTC:
> I am confused about how this problem works. Do we need new a issue or add some message about it?

https://github.com/chaos-mesh/chaosd/blob/6cedaf5f191d8fc25eec95c16309c17324b9c55d/pkg/server/chaosd/stress_attack.go#L119

when recover stress chaos, chaosd will kill the stress-ng process. But `kill` will only send a signal to stress-ng, will not wait for the process exit. Maybe stress-ng exit cost more than 1 second, and then the test failed.

And I print the stress-ng's process info when test failed @Andrewmatilde

test/integration_test/stress/run.sh
```diff
@@ -32,11 +32,11 @@ fi
 uid=`cat cpu.out | grep "Attack stress cpu successfully" | awk -F: '{print $2}'`
 ${bin_path}/chaosd recover ${uid}
 
-sleep 1
+sleep 2
```
Andrewmatilde, line 35: How about wait for process here?

cwen0: Agree with this

WangXiangUSTC: good suggestion, fixed, PTAL again

Andrewmatilde: LGTM

cwen0 (approved):
LGTM

PR #42, opened 2021-03-29
add percentage of disk option in disk fill

Andrewmatilde:
Usage:
```
fill disk

Usage:

Flags:
  -f, --fallocate        fill disk by fallocate instead of dd (default true)
  -h, --help             help for fill
  -p, --path string      'path' specifies the location to fill data in.
                         If path not provided, a temp file will be generated and deleted immediately after data filled in or allocated
  -c, --percent string   'percent' how many percent data of disk will fill in the file path
  -s, --size string      'size' specifies how many data will fill in the file path with unit MB.

Global Flags:
```


Andrewmatilde:
@WangXiangUSTC PTAL

pkg/core/disk.go
```diff
@@ -23,8 +25,9 @@ const (
 
 type DiskCommand struct {
 	Action          string `json:"action"`
-	Size            uint64 `json:"size"`
+	Size            string `json:"size"`
```
WangXiangUSTC, line 28: why change the size from int to string?

pkg/server/chaosd/disk.go
```diff
@@ -149,12 +151,37 @@ func (s *Server) DiskFill(fill *core.DiskCommand) (uid string, err error) {
...
+		di, err := disk.GetInfo(dir)
+		if err != nil {
+			log.Error(fmt.Sprintf("unexpected err when using disk.GetInfo"), zap.Error(err))
+			return uid, err
+		}
+		totalM := di.Total / 1024 / 1024
+		fill.Size = strconv.FormatUint(totalM*percent/100, 10)
+	} else {
+		fill.Size = "0"
+	}
```
WangXiangUSTC, line 177: how about do some validate work in `func (d *DiskCommand) Validate() error`, for example, at least one of `Size` and `Percent` can not be empty.

Andrewmatilde: I agree with you.

pkg/core/disk.go
Andrewmatilde: Because size will read from string, and used as string (sprint). The another reason is I want a empty string as default value means None size , but not a 0 value in u64.

WangXiangUSTC: OK

PR #46, opened 2021-04-09
add percentage of disk option in disk fill

Andrewmatilde:
Usage:
```
fill disk

Usage:

Flags:
  -f, --fallocate        fill disk by fallocate instead of dd (default true)
  -h, --help             help for fill
  -p, --path string      'path' specifies the location to fill data in.
                         If path not provided, a temp file will be generated and deleted immediately after data filled in or allocated
  -c, --percent string   'percent' how many percent data of disk will fill in the file path
  -s, --size string      'size' specifies how many data will fill in the file path with unit MB.

Global Flags:
```


Andrewmatilde:
@WangXiangUSTC PTAL

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
@cwen0 PTAL

WangXiangUSTC:
this pr is duplicate with https://github.com/chaos-mesh/chaosd/pull/42, and have some comment in it

pkg/server/chaosd/disk.go
```diff
@@ -150,11 +152,34 @@ func (s *Server) DiskFill(fill *core.DiskCommand) (uid string, err error) {
...
+	if fill.Size != "" {
+		fill.Size = strings.Trim(fill.Size, " ")
+	} else if fill.Percent != "" {
+		fill.Percent = strings.Trim(fill.Percent, " ")
+		percent, err := strconv.ParseUint(fill.Percent, 10, 0)
+		if err != nil {
+			log.Error(fmt.Sprintf(" unexcepted err when parsing disk percent '%s'", fill.Percent), zap.Error(err))
+			return uid, err
+		}
+		dir, err := os.Getwd()
```
cwen0, line 164: Do we need to consider `file.Path` here?

WangXiangUSTC: do you mean `fill.Path`?

cwen0: Yes, typo has been fixed 😂

Andrewmatilde: Ok, I agree with you.

Andrewmatilde:
PTAL, @WangXiangUSTC , @cwen0

pkg/server/chaosd/disk.go
```diff
@@ -124,17 +125,15 @@ func (diskAttack) attackDiskFill(fill *core.DiskCommand) error {
 			log.Error(fmt.Sprintf(" unexcepted err when parsing disk percent '%s'", fill.Percent), zap.Error(err))
 			return err
 		}
-		dir, err := os.Getwd()
+		dir := filepath.Dir(fill.Path)
+		s := syscall.Statfs_t{}
```
WangXiangUSTC, line 129: why use this instead of `disk.GetInfo`?

Andrewmatilde: This is just the implements of disk.GetInfo. I change it mostly because minio is hard to import into go mod , and pass to ci.And minio is also not recommand to be used as a lib.

WangXiangUSTC (approved):
LGTM

cwen0 (approved):
LGTM

PR #31, opened 2021-03-16
Enhance HTTP server & some refactors

shivanshs9:
- Refactored server and cmd components, simplifying the process of attack.
- Refactored all existing attack types to use the new convention of implementing attack. 
- Added ExperimentRun model to keep track of each run of an experiment. In the case of direct attack, there will only be one run. In the case of scheduling attack, there will be multiple runs corresponding to each scheduled attack.
- API: Added system endpoints for health check and version info.
- API: Added endpoint to list experiments.

PR #50, opened 2021-04-14
Implement Scheduling

shivanshs9:
TODO: Add unit tests for scheduler


pkg/core/experiment.go
```diff
@@ -58,5 +61,33 @@ type Experiment struct {
...
+	cachedRequestCommand AttackConfig
+}
+
+func (exp *Experiment) GetRequestCommand() (AttackConfig, error) {
+	if exp.cachedRequestCommand != nil {
+		return exp.cachedRequestCommand, nil
+	}
+
+	var attackConfig AttackConfig
+	switch exp.Kind {
```
AsterNighT, line 73: I would suggest put this in a separate `func getCommandByKind(kind string) AttackConfig` and put it in some `utils.go` for reuse.

pkg/server/chaosd/recover.go
```diff
@@ -61,6 +61,7 @@ func (s *Server) RecoverAttack(uid string) error {
 	}
 
 	env := s.newEnvironment(uid)
+	// TODO: In case a scheduled cron is running right now, recovery may lead system to an unknown state
```
AsterNighT, line 64: What about a `pending` state indicating the cron is running now? Or simply give it a lock.

AsterNighT (commented):
Rest lgtm

pkg/core/experiment_run.go
```diff
@@ -24,17 +32,27 @@ type ExperimentRunStore interface {
...
+	Update(ctx context.Context, runUid string, status string, message string) error
 	// TODO: experiment run creation and updation
 }
 
 // ExperimentRun represents a run of an experiment
 type ExperimentRun struct {
 	ID           uint      `gorm:"primary_key" json:"id"`
 	UID          string    `gorm:"index:uid" json:"uid"`
-	StartAt      time.Time `json:"start_at"`
+	StartAt      time.Time `gorm:"autoCreateTime" json:"start_at"`
```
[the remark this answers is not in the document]
shivanshs9: with that gorm name, it automatically sets the creation timestamp on insertion.

pkg/core/experiment.go
shivanshs9: @AsterNighT hmmm.. the use case is unmarshalling here so using interface `AttackConfig` works. In other use-cases, I can't imagine a utility function which just returns an interface, and without even deserialized data, would be reusable. Since we'd still need to do this switch case again to differentiate between which of the concrete types the interface belongs to.

pkg/server/chaosd/recover.go
shivanshs9: So experiment will have `Scheduled` state once created. But on each job run, it'll switch experiment's state to `Pending` and back to `Scheduled` once done?

shivanshs9: Lock would've been generally a good idea but here recover picks up exp data from DB. I could save running jobs in a map, but not sure how good of an idea that is. :thinking: 
I am thinking to instead check for the last run here and if it's in `RunStarted` state, then don't recover and throw an error. But even with this, it's definitely possible that both recovery and execution happens due to multithreaded server.

pkg/scheduler/cron.go
```diff
@@ -14,45 +14,103 @@
...
-// TODO: on running, create new experiment run
 // TODO: write tests for it
-func (cj CronJob) Run() {
-	cj.run()
+func (cj *CronJob) Run() {
+	log.Info("Started new run", zap.String("expId", cj.experiment.Uid))
+	var err error
+	cfg, err := cj.experiment.GetRequestCommand()
+	if err != nil {
+		panic(perr.WithStack(err))
```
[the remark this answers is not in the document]
shivanshs9: @WangXiangUSTC Fixed now by recovering panics in defer call above.

pkg/server/chaosd/recover.go
AsterNighT: A similar approach (use a map to store jobs and locks) is adopted in `chaos-mesh/pkg/bpm` https://github.com/chaos-mesh/chaos-mesh/blob/8ad8a6ff66abbceb0707e89e2169b08e8b7692fa/pkg/bpm/bpm.go#L83. I cannot say it's good or bad but you can take a look at it.

shivanshs9: will check this implementation

WangXiangUSTC:
@shivanshs9 could you please add the usage of the scheduler in the pr's description,   demo is OK

WangXiangUSTC:
I use `curl -X POST 127.0.0.1:31767/api/attack/stress -H "Content-Type:application/json" -d '{"action":"mem", "workers": 2, "size": "100m", "schedule": "*/1 * * * *", "duration": "30s"}'` to create experiment, and chaosd report error when run it:

```log
[cron.go:52] ["scheduled run errored"] [expId=d41aa6e1-e869-4abe-86c8-20a1a01165ca] [error="json: cannot unmarshal number into Go struct field ProcessCommand.duration of type string"] [errorVerbose="json: cannot unmarshal number into Go struct field ProcessCommand.duration of type string
```

And then I try to delete it by `curl -X DELETE 127.0.0.1:31767/api/attack/d41aa6e1-e869-4abe-86c8-20a1a01165ca`, and it can't be delete:

```log
{"error":true,"message":"error.api.internal_server_error: Recover experiment d41aa6e1-e869-4abe-86c8-20a1a01165ca failed: json: cannot unmarshal number into Go struct field ProcessCommand.duration of type string","code":"error.api.internal_server_error","full_text":"error.api.internal_server_error: Recover experiment d41aa6e1-e869-4abe-86c8-20a1a01165ca failed: json: cannot unmarshal number into Go struct field ProcessCommand.duration of type string...}
```

shivanshs9:
@WangXiangUSTC that was due to a bug I found in #31. It was my oversight to miss it back then but now I fixed it here. Basically, in the DB the stress attack was being stored as process attack struct.

shivanshs9:
@WangXiangUSTC I also found this duplicate `Duration` field in **StressCommand**:
https://github.com/shivanshs9/chaosd/blob/a5e2b508bfe32051e0f2c7d2ce94b93b886110a6/pkg/core/stress.go#L35

I don't think it causes any issues in runtime right now because it's a nested struct + json key is different in case.
However, it may very well cause confusion later due to two similar keys in the marshalled json. I guess I should rename the `Duration` field in `SchedulerConfig` to something like `ScheduleDuration`? Or do you recommend just changing the json name of the field??

WangXiangUSTC:
> @WangXiangUSTC I also found this duplicate `Duration` field in **StressCommand**:
> https://github.com/shivanshs9/chaosd/blob/a5e2b508bfe32051e0f2c7d2ce94b93b886110a6/pkg/core/stress.go#L35
> 
> I don't think it causes any issues in runtime right now because it's a nested struct + json key is different in case.
> However, it may very well cause confusion later due to two similar keys in the marshalled json. I guess I should rename the `Duration` field in `SchedulerConfig` to something like `ScheduleDuration`? Or do you recommend just changing the json name of the field??

I think we can delete the `Duration` in `StressCommand`, it is useless now

pkg/scheduler/cron.go
```diff
@@ -14,45 +14,129 @@
...
+	log.Info("Started new run", zap.String("expId", cj.experiment.Uid))
+	cfg, err := cj.experiment.GetRequestCommand()
+	if err != nil {
+		panic(perr.WithStack(err))
+	}
+	cronDuration, err := cfg.ScheduleDuration()
+	if err != nil {
+		panic(perr.WithStack(err))
+	}
+	if cronDuration != nil && hasCronDurationExceeded(cj.experiment.CreatedAt, *cronDuration) {
```
WangXiangUSTC, line 83: seems that the experiment can only run one time. For example, the `schedule` is `*/1 * * * *`, and the `duration` is `30s`, the experiment will run at 17:01:00, 17:02:00 ..., and will keep 30s every time, and then recover it.

but the `hasCronDurationExceeded` here will return true when last 30s, and then remove it.

shivanshs9: @WangXiangUSTC I tested it but it seems to run fine for me :thinking: 
```
[2021/05/07 13:45:50.671 +05:30] [INFO] [zapr.go:69] [added] [now=2021/05/07 08:15:50.671 +00:00] [entry=1] [next=2021/05/07 08:16:00.000 +00:00]
[GIN] 2021/05/07 - 13:45:50 | 200 |  161.479713ms |       127.0.0.1 | POST     "/api/attack/stress"
[2021/05/07 13:46:00.002 +05:30] [INFO] [zapr.go:69] [wake] [now=2021/05/07 08:16:00.001 +00:00]
[2021/05/07 13:46:00.002 +05:30] [INFO] [zapr.go:69] [run] [now=2021/05/07 08:16:00.001 +00:00] [entry=1] [next=2021/05/07 08:17:00.000 +00:00]
[2021/05/07 13:46:00.002 +05:30] [INFO] [cron.go:74] ["Started new run"] [expId=54d519bf-5458-4f2e-b811-ab0d8a1e71a6]
[2021/05/07 13:46:00.013 +05:30] [INFO] [cron.go:98] ["executing attack on new exp run"] [expRunUID=f7ea65e5-e506-4182-afd7-bae5f89ee4b1]
[2021/05/07 13:46:00.013 +05:30] [INFO] [stress.go:65] ["stressors normalize"] [arguments=" --vm 2 --vm-keep --vm-bytes 100000000"]
[2021/05/07 13:47:00.000 +05:30] [INFO] [zapr.go:69] [wake] [now=2021/05/07 08:17:00.000 +00:00]
[2021/05/07 13:47:00.000 +05:30] [INFO] [zapr.go:69] [run] [now=2021/05/07 08:17:00.000 +00:00] [entry=1] [next=2021/05/07 08:18:00.000 +00:00]
[2021/05/07 13:47:00.001 +05:30] [INFO] [cron.go:74] ["Started new run"] [expId=54d519bf-5458-4f2e-b811-ab0d8a1e71a6]
[2021/05/07 13:47:00.001 +05:30] [INFO] [zapr.go:69] [removed] [entry=1]
[2021/05/07 13:47:00.010 +05:30] [INFO] [cron.go:62] ["scheduled run success"] [expId=54d519bf-5458-4f2e-b811-ab0d8a1e71a6]
[2021/05/07 13:47:00.010 +05:30] [INFO] [cron.go:66] ["cron run finished"]
```

^ as seen in logs, it ran only once since the first run was only 10s after scheduling it.

pkg/scheduler/cron.go
```diff
@@ -14,45 +14,168 @@
...
+	cj.lock.Lock()
+	cj.waitForRecovery = waitForRecovery
+	cj.lock.Unlock()
 }
 
-// TODO: on running, create new experiment run
 // TODO: write tests for it
-func (cj CronJob) Run() {
-	cj.run()
+func (cj *CronJob) Run() {
```
WangXiangUSTC, line 78: seems that the `CronJob` is implementing the interface [Job](https://github.com/robfig/cron/blob/bc59245fe10efaed9d51b56900192527ed733435/cron.go#L35) how about adding a comment?

shivanshs9: sure, will add it

pkg/scheduler/cron.go
```diff
@@ -14,45 +14,168 @@
...
+	if err != nil {
+		panic(perr.WithStack(err))
+	}
+
+	newRun = cj.experiment.NewRun()
+	if err = cj.scheduler.expRunStore.NewRun(context.Background(), newRun); err != nil {
+		panic(perr.WithStack(err))
+	}
+
+	if cronDuration != nil {
```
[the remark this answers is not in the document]
shivanshs9: if the duration is nil, the attack will periodically execute as per the schedule, but they won't be recovered. I'm not sure if that's the intended behavior though :thinking:

pkg/store/experiment/experiment_run.go
```diff
@@ -50,7 +50,8 @@ func (store *experimentRunStore) ListByExperimentID(ctx context.Context, id uint
 func (store *experimentRunStore) ListByExperimentUID(ctx context.Context, uid string) ([]*core.ExperimentRun, error) {
 	runs := make([]*core.ExperimentRun, 0)
 	if err := store.db.
-		Joins("JOIN experiments ON experiments.uid = ?", uid).
+		Joins("Experiment").
+		Where("experiment.uid = ?", uid).
```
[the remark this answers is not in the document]
shivanshs9: it apparently wasn't working as expected.. no idea why. it was not filtering based on uid and listing all instead

pkg/scheduler/cron.go
shivanshs9: fixed

pkg/scheduler/cron.go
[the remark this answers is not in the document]
shivanshs9: yep added a common validation

pkg/server/chaosd/recover.go
shivanshs9: redundant now since recovery for runs happens after given duration

pkg/core/common.go
```diff
@@ -54,3 +69,13 @@ func (config CommonAttackConfig) AttackKind() string {
 
 // CompleteDefaults no-op implementation
 func (config *CommonAttackConfig) CompleteDefaults() {}
+
+// Validate does a basic validation of common parameters
+func (config *CommonAttackConfig) Validate() error {
+	if len(config.Schedule) > 0 {
+		if dur, err := config.ScheduleDuration(); dur == nil || err != nil {
+			return errors.New("Provide a valid duration for the scheduled attack")
```
WangXiangUSTC, line 77:
```suggestion
			return errors.New("Provide a invalid duration for the scheduled attack")
```

pkg/scheduler/cron.go
```diff
@@ -14,45 +14,168 @@
...
+	if !cj.assertNoRecovery() {
+		log.Info("skipping scheduled execution of attack since recovery in progress", zap.String("expId", cj.experiment.Uid))
+		return
+	}
+
+	var newRun *core.ExperimentRun
+	var recoverTimer *time.Timer
+	defer func() {
+		var updErr error
+		if panicErr, ok := recover().(error); panicErr != nil && ok {
```
WangXiangUSTC, line 88: I run a scheduler job on stress attack, and the log print `scheduled run success`, but I didn't find the process of stress-ng, at last I find it is because some function in macOS is not implemented.

seems that here should change to `if panicErr := recover(); panicErr != nil {`

pkg/core/common.go
shivanshs9: @WangXiangUSTC it should be "valid" only. The error states what the user should do to fix the error.

WangXiangUSTC: OK, got it

pkg/scheduler/cron.go
shivanshs9: @WangXiangUSTC not sure what u mean, the error is shown in the logs: 
```
[2021/05/17 12:09:00.010 +05:30] [ERROR] [cron.go:89] ["scheduled run errored"] [expId=1c421bdc-5841-4849-ac78-86a6637fcc7c] [error="attack failed: exec: \"stress-ng\": executable file not found in $PATH"] [errorVerbose="exec: \"stress-ng\": executable file not found in $PATH\nattack failed"] [stack="github.com/pingcap/log.Error\n\t/home/shivanshs9/go/pkg/mod/github.com/pingcap/log@v0.0.0-20200117041106-d28c14d3b1cd/global.go:42\ngithub.com/chaos-mesh/chaosd/pkg/scheduler.(*CronJob).Run.func1\n\t/home/shivanshs9/projects/chaos/chaosd/pkg/scheduler/cron.go:89\nruntime.gopanic\n\t/usr/lib/go/src/runtime/panic.go:965\ngithub.com/chaos-mesh/chaosd/pkg/scheduler.(*CronJob).Run\n\t/home/shivanshs9/projects/chaos/chaosd/pkg/scheduler/cron.go:137\ngithub.com/robfig/cron/v3.SkipIfStillRunning.func1.1\n\t/home/shivanshs9/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/chain.go:85\ngithub.com/robfig/cron/v3.FuncJob.Run\n\t/home/shivanshs9/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:136\ngithub.com/robfig/cron/v3.(*Cron).startJob.func1\n\t/home/shivanshs9/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:312"]
```

and also updated in DB:
```
[{"id":296,"uid":"2fcb6147-10a8-4f7f-8675-7af19192a687","start_at":"2021-05-17T12:09:00.0022572+05:30","finished_at":"0001-01-01T00:00:00Z","status":"failed","error":"attack failed: exec: \"stress-ng\": executable file not found in $PATH","ExperimentID":62,"experiment":{"id":62,"uid":"1c421bdc-5841-4849-ac78-86a6637fcc7c","created_at":"2021-05-17T12:08:33.237843307+05:30","updated_at":"2021-05-17T12:08:33.249153029+05:30","status":"scheduled","error":"","kind":"stress","action":"mem","recover_command":"{\"schedule\":\"*/1 * * * *\",\"duration\":\"30s\",\"action\":\"mem\",\"kind\":\"stress\",\"Load\":0,\"Workers\":2,\"Size\":\"100m\",\"Options\":null,\"StressngPid\":0}","launch_mode":"svr"}},{"id":297,"uid":"fc00dc03-c299-4e0f-8c31-e07608fa8874","start_at":"2021-05-17T12:10:00.001614857+05:30","finished_at":"0001-01-01T00:00:00Z","status":"failed","error":"attack failed: exec: \"stress-ng\": executable file not found in $PATH","ExperimentID":62,"experiment":{"id":62,"uid":"1c421bdc-5841-4849-ac78-86a6637fcc7c","created_at":"2021-05-17T12:08:33.237843307+05:30","updated_at":"2021-05-17T12:08:33.249153029+05:30","status":"scheduled","error":"","kind":"stress","action":"mem","recover_command":"{\"schedule\":\"*/1 * * * *\",\"duration\":\"30s\",\"action\":\"mem\",\"kind\":\"stress\",\"Load\":0,\"Workers\":2,\"Size\":\"100m\",\"Options\":null,\"StressngPid\":0}","launch_mode":"svr"}}]
```

WangXiangUSTC: I mean it can't handle the panic error, you can test it by updating one `Attack` function to panic, and check it

shivanshs9: @WangXiangUSTC ah got it, you're referring to this: https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/bpm/build_darwin.go

Technically the issue is that the panic doesn't throw an `error` instance, but I understand changing it would block this PR. No problem, I would handle both error and string in recover

shivanshs9: @WangXiangUSTC fixed now

WangXiangUSTC:
LGTM, thanks @shivanshs9, and maybe we need to add some test for it some days later

WangXiangUSTC (approved):

Andrewmatilde (approved):
LGTM

Issue #41, opened 2021-03-25
support DNS type under Network attack 


WangXiangUSTC:
This feature maybe includes:
- [x] update the address of the DNS server to a wrong value 
- [ ] map a host with a wrong address

This can be implement by modify the file `/etc/hosts` and `/etc/resolv.conf`

AsterNighT:
A typo, s/suport/support/

WangXiangUSTC:
> A typo, s/suport/support/

the title updated 😄

PR #51, opened 2021-04-15
support DNS chaos

WangXiangUSTC:

fix https://github.com/chaos-mesh/chaosd/issues/41, this pr just support update the dns server in /etc/resolv.conf

TODO: map specified host to specified IP

```bash
./bin/chaosd attack network dns --help
attack DNS server

Usage:
  chaosd attack network dns [flags]

Flags:
      --dns-server string   update the DNS server in /etc/resolv.conf with this value (default "123.123.123.123")
  -h, --help                help for dns

Global Flags:
```


pkg/core/network.go
```diff
@@ -141,6 +151,12 @@ func (n *NetworkCommand) setDefaultForNetworkLoss() {
 	}
 }
 
+func (n *NetworkCommand) setDefaultForNetworkDNS() {
```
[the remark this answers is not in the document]
WangXiangUSTC: @muse-dev ignore

pkg/server/chaosd/network.go
```diff
@@ -36,24 +36,31 @@ func (networkAttack) Attack(options core.AttackConfig, env Environment) (err err
 		ipsetName string
 	)
 
-	if attack.NeedApplyIPSet() {
-		ipsetName, err = env.Chaos.applyIPSet(attack, env.AttackUid)
-		if err != nil {
-			return errors.WithStack(err)
+	switch attack.Action {
+	case core.NetworkDNSAction:
```
AsterNighT, line 40: That seems a bit weird. Why can't we use some `if attack.NeedApplyDNS()`?

AsterNighT: But certainly `DNSAttack` will have `attack.NeedApplyIpset() == false`, so the original way would work properly?

WangXiangUSTC: Yes, it can work, but I think use switch can make the logic more clear

AsterNighT: That's ok. Rest lgtm

AsterNighT (approved):

Andrewmatilde (approved):
LGTM

PR #52, opened 2021-04-16
[Discussion] Reorganize	structure of ./cmd to make it clear and reduce the prefix length of its internal package

Andrewmatilde:
The structure of ./cmd now does not meet my aesthetic. It maybe nesting too deeply but do not splite the package well.
So I suggest to reorganize structure of ./cmd.
Before:
2021-04-16 11-45-24 的屏幕截图
After:
2021-04-16 11-46-44 的屏幕截图


Andrewmatilde:
PTAL , @cwen0 , @WangXiangUSTC , @shivanshs9

shivanshs9 (approved):
Indeed, it's a welcome change! :raised_hands: 
LGMT

WangXiangUSTC (approved):
LGTM

Issue #48, opened 2021-04-14
ignore useless log of  go.uber.org/fx with command mode


WangXiangUSTC:
since pr #31 merged, there are many logs of  `go.uber.org/fx` when run with command mode:

```bash
./bin/chaosd attack disk fill -p /tmp/temp -s 100
2021/04/14 02:19:58 [Fx] PROVIDE        *config.Config <= github.com/chaos-mesh/chaosd/cmd/chaosd/ctl/command.glob..func1()
2021/04/14 02:19:58 [Fx] PROVIDE        *dbstore.DB <= github.com/chaos-mesh/chaosd/pkg/store/dbstore.NewDBStore()
2021/04/14 02:19:58 [Fx] PROVIDE        core.ExperimentStore <= github.com/chaos-mesh/chaosd/pkg/store/experiment.NewStore()
2021/04/14 02:19:58 [Fx] PROVIDE        core.ExperimentRunStore <= github.com/chaos-mesh/chaosd/pkg/store/experiment.NewRunStore()
2021/04/14 02:19:58 [Fx] PROVIDE        core.IPSetRuleStore <= github.com/chaos-mesh/chaosd/pkg/store/network.NewIPSetRuleStore()
2021/04/14 02:19:58 [Fx] PROVIDE        core.IptablesRuleStore <= github.com/chaos-mesh/chaosd/pkg/store/network.NewIptablesRuleStore()
2021/04/14 02:19:58 [Fx] PROVIDE        core.TCRuleStore <= github.com/chaos-mesh/chaosd/pkg/store/network.NewTCRuleStore()
2021/04/14 02:19:58 [Fx] PROVIDE        *chaosd.Server <= github.com/chaos-mesh/chaosd/pkg/server/chaosd.NewServer()
2021/04/14 02:19:58 [Fx] PROVIDE        *httpserver.httpServer <= github.com/chaos-mesh/chaosd/pkg/server/httpserver.NewServer()
2021/04/14 02:19:58 [Fx] PROVIDE        crclients.ContainerRuntimeInfoClient <= github.com/chaos-mesh/chaosd/pkg/crclient.NewNodeCRClient()
2021/04/14 02:19:58 [Fx] PROVIDE        int <= os.Getpid()
2021/04/14 02:19:58 [Fx] PROVIDE        *chaosdaemon.DaemonServer <= github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon.NewDaemonServerWithCRClient()
[...]

PR #59, opened 2021-04-22
ignore fx log

WangXiangUSTC:

fix issue https://github.com/chaos-mesh/chaosd/issues/48, ignore fx logs.


WangXiangUSTC:
@Andrewmatilde @shivanshs9 PTAL

shivanshs9:
@WangXiangUSTC I think we can configure their log level to DEBUG instead? :thinking:

WangXiangUSTC:
> @WangXiangUSTC I think we can configure their log level to DEBUG instead? 🤔

Seems that DEBUG will print more log 🤔

WangXiangUSTC:
> @WangXiangUSTC I think we can configure their log level to DEBUG instead? 🤔

Sorry, I make a mistake, I edit your comment, in fact I want to reply it 🤣

WangXiangUSTC:
only print log of go.uber.org/fx in debug mode now, PTAL again @Andrewmatilde @shivanshs9

shivanshs9 (approved):
LGTM!

cwen0 (approved):
LGTM

PR #60, opened 2021-04-22
Add more disk attack support

Andrewmatilde:
1. Add support of unit for disk attack fill & payload.
2. Add support for user to control if destroy file created in disk filling.
3. Put empty string "" in path of disk read&write payload will now create an empty temp file to write , add read from device path of "/".
4. Multi process is now supported

- [x] Unit test


Andrewmatilde:
PTAL, @WangXiangUSTC , @cwen0

pkg/server/chaosd/disk.go
```diff
@@ -38,84 +37,96 @@ const DDWritePayloadCommand = "dd if=/dev/zero of=%s bs=%s count=%s oflag=dsync"
...
-			err := errors.Errorf("empty read payload path")
-			log.Error(fmt.Sprintf("payload action: %s", payload.Action), zap.Error(err))
-			return err
+			path, err := utils.GetRootDevice()
+			if err != nil {
+				log.Error("err when GetRootDevice in reading payload", zap.Error(err))
+				return err
+			}
+			if path == "" {
+				err = fmt.Errorf("can not get root device path")
```
WangXiangUSTC, line 76: here use `fmt.Errorf`, but L82 use `errors.Errorf`, it's better to unify them, only use one

pkg/server/chaosd/disk.go
```diff
@@ -38,84 +37,96 @@ const DDWritePayloadCommand = "dd if=/dev/zero of=%s bs=%s count=%s oflag=dsync"
...
+	}
+
+	byteSize, err := utils.ParseUnit(payload.Size + payload.Unit)
+	if err != nil {
+		log.Error(fmt.Sprintf("fail to get parse size unit , %s, %s", payload.Size, payload.Unit), zap.Error(err))
+		return err
+	}
+	sizeBlocks := utils.SplitByteSize(byteSize, payload.PayloadProcessNum)
+
+	for _, sizeBlock := range sizeBlocks {
```
WangXiangUSTC, line 94: should here use goroutine to run dd parallel？

cmd/attack/disk.go
```diff
@@ -70,14 +70,19 @@ func NewDiskWritePayloadCommand(dep fx.Option, options *core.DiskCommand) *cobra
 	}
 
 	cmd.Flags().StringVarP(&options.Size, "size", "s", "",
-		"'size' specifies how many data will fill in the file path with unit MB.")
+		"'size' specifies how many data will fill in the file path with unit.")
 	cmd.Flags().StringVarP(&options.Path, "path", "p", "/dev/null",
 		"'path' specifies the location to fill data in.\n"+
 			"If path not provided, payload will write into /dev/null")
+	cmd.Flags().StringVarP(&options.Unit, "unit", "u", "M",
```
WangXiangUSTC, line 77: can we combine the `size` and `unit` to only one config, just like the `size` in stress attack:

```
-s, --size string       Size specifies N bytes consumed per vm worker, default is the total available memory. One can specify the size as % of total available memory or in units of B, KB/KiB, MB/MiB, GB/GiB, TB/TiB..
```

pkg/utils/disk_darwin.go
```diff
@@ -28,3 +28,7 @@ func GetDiskTotalSize(path string) (total uint64, err error) {
 	total = uint64(s.Bsize) * (s.Blocks - reservedBlocks)
 	return total, err
 }
+
+func GetRootDevice() (string, error) {
```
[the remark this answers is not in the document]
Andrewmatilde: OK，I aggree with you.

pkg/server/chaosd/disk.go
Andrewmatilde: OK，I aggree with you.

pkg/server/chaosd/disk.go
Andrewmatilde: OK，I aggree with you.

cmd/attack/disk.go
Andrewmatilde: OK，I aggree with you.

pkg/server/chaosd/disk.go
```diff
@@ -38,84 +38,118 @@ const DDWritePayloadCommand = "dd if=/dev/zero of=%s bs=%s count=%s oflag=dsync"
...
+				log.Error(string(output), zap.Error(err))
+				fatalErrors <- err
+			}
+			log.Info(string(output))
+			wg.Done()
+		}(cmd)
+	}
+
+	go func() {
+		wg.Wait()
```
WangXiangUSTC, line 116: why use `wait` in goroutine?

Andrewmatilde: wg.Wait() will wait blocks until the wg counter is zero. wg.Add(len(sizeBlocks)) will add len(sizeBlocks) to wg, and wg.Done is wg.Add(-1).

pkg/server/chaosd/disk.go
```diff
@@ -35,87 +36,132 @@ type diskAttack struct{}
 var DiskAttack AttackType = diskAttack{}
 
 const DDWritePayloadCommand = "dd if=/dev/zero of=%s bs=%s count=%s oflag=dsync"
-const DDReadPayloadCommand = "dd if=%s of=/dev/null bs=%s count=%s iflag=dsync,direct,fullblock"
+const DDReadPayloadCommand = "dd if=%s of=/dev/null bs=%s count=%s iflag=dsync, fullblock, nocache"
```
WangXiangUSTC, line 39: why need to update it?

pkg/server/chaosd/disk.go
```diff
@@ -35,87 +36,132 @@ type diskAttack struct{}
...
+		cmd := exec.Command("bash", "-c", fmt.Sprintf(cmdFormat, payload.Path, sizeBlock.BlockSize, sizeBlock.Size))
+
+		go func(cmd *exec.Cmd) {
+			if locker != 0 {
+				return
+			}
+			output, err := cmd.CombinedOutput()
+			if err != nil {
+				log.Info(cmd.String())
+				log.Error(string(output), zap.Error(err))
```
WangXiangUSTC, line 115: how about combining these two logs into one?

WangXiangUSTC (commented):
rest LGTM

pkg/server/chaosd/disk.go
Andrewmatilde: OK.

pkg/server/chaosd/disk.go
Andrewmatilde: Direct iflag will forbid user from reading virtual device, so I use nocache to instead it which is  similar with direct.

pkg/server/chaosd/disk.go
```diff
@@ -35,87 +36,138 @@ type diskAttack struct{}
...
+	var wg sync.WaitGroup
+	var locker uint32
+	wg.Add(len(sizeBlocks))
+	fatalErrors := make(chan error)
+	wgDone := make(chan bool)
+	for _, sizeBlock := range sizeBlocks {
+		cmd := exec.Command("bash", "-c", fmt.Sprintf(cmdFormat, payload.Path, sizeBlock.BlockSize, sizeBlock.Size))
+
+		go func(cmd *exec.Cmd) {
+			if locker != 0 {
```
WangXiangUSTC, line 116: should use atomic to read

Andrewmatilde:
PTAL @WangXiangUSTC

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
@cwen0 PTAL

Andrewmatilde:
PTAL , @longfangsong

cmd/attack/disk.go
```diff
@@ -70,14 +70,19 @@ func NewDiskWritePayloadCommand(dep fx.Option, options *core.DiskCommand) *cobra
 	}
 
 	cmd.Flags().StringVarP(&options.Size, "size", "s", "",
-		"'size' specifies how many data will fill in the file path with unit MB.")
-	cmd.Flags().StringVarP(&options.Path, "path", "p", "/dev/null",
+		"'size' specifies how many units of data will write into the file path.\n"+
+			"'unit' specifies the unit of data, support c=1, w=2, b=512, kB=1000, K=1024, MB=1000*1000,"+
+			"M=1024*1024, , GB=1000*1000*1000, G=1024*1024*1024 BYTES\n"+
+			"example : 1M | 512kB")
```
longfangsong, line 76: You can use `` to create multiline string in Golang.

Andrewmatilde: Thx, I find description without newline seems better.

cmd/attack/disk_test.go
```diff
@@ -15,20 +15,21 @@ package attack
 
 import (
 	"os"
-	"strconv"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
cwen0, line 20: Format imports

pkg/server/chaosd/disk.go
```diff
@@ -35,87 +36,135 @@ type diskAttack struct{}
...
-	return disk.attackDiskPayload(attack)
+	return disk.diskPayload(attack)
 }
 
-func (diskAttack) attackDiskPayload(payload *core.DiskCommand) error {
+// diskPayload will execute a dd command (DDWritePayloadCommand or DDReadPayloadCommand)
+// to add a write or read payload.
+func (diskAttack) diskPayload(payload *core.DiskOption) error {
+	var cmdFormat string
 	switch payload.Action {
```
cwen0, line 54: Can we abstract this logic to a single function? such as `initPlayloadPath` or others.

cwen0 (commented):
Rest LGTM

pkg/server/chaosd/disk.go
Andrewmatilde: OK, I agree with you.

Andrewmatilde: Done. But I could not abstract the logic that temp file of wirte payload should be ```defer remove``` in the diskPayload func.

cwen0 (approved):
LGTM

PR #69, opened 2021-05-24
feat: support ssl + client ssl authentication

shivanshs9:
Implements Authentication and SSL support in Chaosd (as per https://github.com/chaos-mesh/rfcs/pull/14)

For implementation, I referred [this guide](https://venilnoronha.io/a-step-by-step-guide-to-mtls-in-go)

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -32,6 +32,9 @@ func NewServerCommand() *cobra.Command {
 
 	cmd.Flags().IntVarP(&conf.ListenPort, "port", "p", 31767, "listen port of the Chaosd Server")
 	cmd.Flags().StringVarP(&conf.ListenHost, "host", "a", "0.0.0.0", "listen host of the Chaosd Server")
+	cmd.Flags().StringVar(&conf.SSLCertFile, "cert", "", "path to a PEM encoded certificate file")
+	cmd.Flags().StringVar(&conf.SSLKeyFile, "key", "", "path to a PEM encoded private key file")
+	cmd.Flags().StringVar(&conf.SSLClientCAFile, "CA", "", "path to a PEM encoded CA's certificate file")
 	cmd.Flags().StringVarP(&conf.Runtime, "runtime", "r", "docker", "current container runtime")
 	cmd.Flags().BoolVar(&conf.EnablePprof, "enable-pprof", true, "enable pprof")
 	cmd.Flags().IntVar(&conf.PprofPort, "pprof-port", 31766, "listen port of the pprof server")
--- a/pkg/server/httpserver/auth.go
+++ b/pkg/server/httpserver/auth.go
@@ -1,0 +1,129 @@
+// Copyright 2021 Chaos Mesh Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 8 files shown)


WangXiangUSTC:
@shivanshs9  Thanks, could you please add an integration test for it, to make sure the SSL authentication can work

shivanshs9:
@WangXiangUSTC yep, I'll add one soon. Btw, I already tested it manually. If you'd like to give it a go too, you can generate certificates with [this script](https://github.com/julie-ng/nodejs-certificate-auth/blob/master/generate-certs.sh).
Then I ran the server with: `./bin/chaosd server --cert ~/packages/nodejs-certificate-auth/server/server_cert.pem --key ~/packages/nodejs-certificate-auth/server/server_key.pem --CA ~/packages/nodejs-certificate-auth/server/server_cert.pem`

And then I did bunch of requests:
- HTTPS with no client certificate (got 401)
- HTTPS with self-signed client certificate (got 403)
- HTTPS with authorized client certificate (got 200)

pkg/server/httpserver/auth.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+
+	"github.com/chaos-mesh/chaosd/pkg/server/utils"
+)
+
+const (
+	MTLSServer = "mTLS"
+	TLSServer  = "tls"
```
WangXiangUSTC, line 32: In which case should we use `TLSServer`/`MTLSServer`?

shivanshs9: I probably should add a comment there :thinking:. But to summarize,
1. **TLS:** Standard SSL encryption on requests with the provided server certificate and key. Basically clients will do HTTPS request without presenting any certificates.
2. **mTLS (or Mutual TLS):** Two-way SSL where both server and client presents their certificates. Basically clients do HTTPS request, validates server certificate and then presents its own client certificate. If the client certificate is valid (if it is signed by the CA certificate provided to chaosd), then the client is authenticated and allowed to call any API endpoint.

WangXiangUSTC: OK, got it

WangXiangUSTC (approved):
LGTM

cmd/server/server.go
```diff
@@ -32,6 +32,9 @@ func NewServerCommand() *cobra.Command {
 
 	cmd.Flags().IntVarP(&conf.ListenPort, "port", "p", 31767, "listen port of the Chaosd Server")
 	cmd.Flags().StringVarP(&conf.ListenHost, "host", "a", "0.0.0.0", "listen host of the Chaosd Server")
+	cmd.Flags().StringVar(&conf.SSLCertFile, "cert", "", "path to a PEM encoded certificate file")
+	cmd.Flags().StringVar(&conf.SSLKeyFile, "key", "", "path to a PEM encoded private key file")
+	cmd.Flags().StringVar(&conf.SSLClientCAFile, "CA", "", "path to a PEM encoded CA's certificate file")
```
iguoyr, line 37: It's better to use lowercase `ca` instead.

Andrewmatilde (approved):
LGTM

cmd/server/server.go
WangXiangUSTC: maybe we can update in another pr

PR #73, opened 2021-06-03
Fix bug and implement recover for disk chaos

Andrewmatilde:

Fix bugs in [SplitBytesByProcessNum](https://github.com/chaos-mesh/chaosd/blob/main/pkg/utils/units.go#L92) that put BlockSize and Count value into wrong place with exchange them.

Add make overwrite in disk fill invalid.
Add recover with remove filled file or wrote file.

### DIFF
--- a/pkg/core/disk.go
+++ b/pkg/core/disk.go
@@ -38,7 +38,6 @@ type DiskOption struct {
 	PayloadProcessNum uint8  `json:"payload_process_num"`
 
 	FillByFallocate bool `json:"fill_by_fallocate"`
-	DestroyFile     bool `json:"destroy_file"`
 }
 
 var _ AttackConfig = &DiskOption{}
@@ -85,7 +84,7 @@ func (d *DiskOption) Validate() error {
 				return err
 			}
 		} else {
-			return fmt.Errorf("write into a existing file when not forcing overwrite")
+			return fmt.Errorf("fill into a existing file")
 		}
 	}
 
--- a/pkg/core/disk.go
+++ b/pkg/core/disk.go
@@ -84,7 +84,7 @@ func (d *DiskOption) Validate() error {
 				return err
 			}
 		} else {
-			return fmt.Errorf("fill into a existing file")
+			return fmt.Errorf("fill into an existing file")
 		}
 	}
 
--- a/pkg/core/disk.go
+++ b/pkg/core/disk.go
@@ -61,8 +61,8 @@ func (d *DiskOption) Validate() error {
 		}
 	}
 
-	if d.Action == DiskFillAction {
-		if d.FillByFallocate && byteSize == 0 {
+	if d.Action == DiskFillAction || d.Action == DiskWritePayloadAction {
+		if d.Action == DiskFillAction && d.FillByFallocate && byteSize == 0 {
 			return fmt.Errorf("fallocate not suppurt 0 size or 0 percent data, "+
 				"if you want allocate a 0 size file please set fallocate=false, DiskOption : %v", d)
 		}
@@ -84,7 +84,10 @@ func (d *DiskOption) Validate() error {
 				return err
 			}
 		} else {
-			return fmt.Errorf("fill into an existing file")
+			if d.Action == DiskFillAction {
+				return fmt.Errorf("fill into an existing file")
+			}
+			return fmt.Errorf("write into an existing file")
 		}
 	}
 
--- a/pkg/core/disk.go
+++ b/pkg/core/disk.go
@@ -16,7 +16,12 @@ package core
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"strconv"
+
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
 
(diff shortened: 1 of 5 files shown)


Andrewmatilde:
LSTM, @WangXiangUSTC

pkg/core/disk.go
```diff
@@ -58,11 +60,32 @@ func (d *DiskOption) Validate() error {
...
+					}
+					if err := os.Remove(d.Path); err != nil {
+						return err
+					}
+				}
+			} else {
+				return err
+			}
+		} else {
+			return fmt.Errorf("fill into a existing file")
```
WangXiangUSTC, line 87:
```suggestion
			return fmt.Errorf("fill into an existing file")
```

pkg/server/chaosd/disk.go
```diff
@@ -157,7 +151,7 @@ func (diskAttack) diskPayload(payload *core.DiskOption) error {
 	return nil
 }
 
-const DDFillCommand = "dd if=/dev/zero of=%s bs=%s count=%s iflag=fullblock"
+const DDFillCommand = "dd if=/dev/zero of=%s bs=%s count=%s iflag=fullblock oflag=append conv=notrunc"
```
WangXiangUSTC, line 154: how about adding a comment

pkg/core/disk.go
```diff
@@ -58,11 +60,32 @@ func (d *DiskOption) Validate() error {
...
 		}
 	}
+
 	if d.Action == DiskFillAction {
 		if d.FillByFallocate && byteSize == 0 {
 			return fmt.Errorf("fallocate not suppurt 0 size or 0 percent data, "+
 				"if you want allocate a 0 size file please set fallocate=false, DiskOption : %v", d)
 		}
+
+		_, err := os.Stat(d.Path)
```
WangXiangUSTC, line 70: do we need to add the same judgment for the write payload?

pkg/core/disk.go
```diff
@@ -58,11 +60,32 @@ func (d *DiskOption) Validate() error {
...
+			if os.IsNotExist(err) {
+				// check if Path of file is valid when Path is not empty
+				if d.Path != "" {
+					var b []byte
+					if err := ioutil.WriteFile(d.Path, b, 0644); err != nil {
+						return err
+					}
+					if err := os.Remove(d.Path); err != nil {
+						return err
+					}
```
WangXiangUSTC, line 81: why write something and then remove it?

pkg/server/chaosd/disk.go
Andrewmatilde: OK

pkg/core/disk.go
Andrewmatilde: OK

pkg/core/disk.go
Andrewmatilde: This is a empirical way to check if a path is valid.
[url](https://stackoverflow.com/questions/35231846/golang-check-if-string-is-valid-path)

WangXiangUSTC: maybe we can ignore check here, and after run with this invalid path, dd/fallocate will  report the error message

Andrewmatilde: I think in future users may combine some attacks with our scheduler, if they apply some attacks like:first -> network attack for a day, next -> disk attack for a day, than check invalid path in validation part will bring convenience to found problem ahead of long time executing. And another problem is the error message of dd or fallocate sometimes may bring confuse to user.

pkg/server/chaosd/disk.go
Andrewmatilde: fixed

pkg/core/disk.go
Andrewmatilde: fixed

pkg/core/disk.go
Andrewmatilde: added

WangXiangUSTC (approved):
LGTM

pkg/server/chaosd/disk.go
```diff
@@ -196,29 +182,56 @@ func (diskAttack) diskFill(fill *core.DiskOption) error {
 			log.Error("fail to get disk total size", zap.Error(err))
 			return err
 		}
-		fill.Size = strconv.FormatUint(totalSize*percent/100, 10)
+		fill.Size = strconv.FormatUint(totalSize*percent/100, 10) + "c"
```
cwen0, line 185: Why need to add `c` here?

Andrewmatilde: `c` is a byte size unit in [dd](https://man7.org/linux/man-pages/man1/dd.1.html) means one byte.
`       N and BYTES may be followed by the following multiplicative
       suffixes: c=1, w=2, b=512, kB=1000, K=1024, MB=1000*1000,
       M=1024*1024, xM=M, GB=1000*1000*1000, G=1024*1024*1024, and so on
       for T, P, E, Z, Y.  Binary prefixes can be used, too: KiB=K,
       MiB=M, and so on.`

cwen0 (approved):
LGTM

Issue #61, opened 2021-04-25
version `GLIBC_2.28' not found


WangXiangUSTC:
```bash
./chaosd --help
./chaosd: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by ./chaosd)
```

WangXiangUSTC:
it is because that github.com/chaos-mesh/chaos-mesh/pkg/ptrace need to use cgo

WangXiangUSTC:
the sqlite also require cgo

WangXiangUSTC:
since pr #74 merged, chaosd require at least v2.17, and it is suitable for most Linux versions.

PR #74, opened 2021-06-10
use glibc 2.17 to build chaosd

WangXiangUSTC:

fix issue https://github.com/chaos-mesh/chaosd/issues/61


.github/workflows/release_latest.yml
```diff
@@ -18,13 +21,26 @@ jobs:
         with:
           # Must use at least depth 2!
           fetch-depth: 2
-      - uses: actions/setup-python@v2
+
+      - name: Setup python3
```
WangXiangUSTC, line 25: seems that use `actions/setup-python@v2` to install python in this container will meet some troubles,  will get error when use pip3 to install awscli, don't know why,
so just install it by yum.

Andrewmatilde (approved):
LGTM

cwen0 (approved):
LGTM

PR #76, opened 2021-06-11
some minor update

WangXiangUSTC:
1. support search all kind experiment
2. remove `workers` config in stress memory, because every worker will cost some cpu load, so just use one worker is OK

### DIFF
--- a/cmd/attack/stress.go
+++ b/cmd/attack/stress.go
@@ -53,6 +53,7 @@ func NewStressCPUCommand(dep fx.Option, options *core.StressCommand) *cobra.Comm
 		Short: "continuously stress CPU out",
 		Run: func(*cobra.Command, []string) {
 			options.Action = core.StressCPUAction
+			options.CompleteDefaults()
 			utils.FxNewAppWithoutLog(dep, fx.Invoke(stressAttackF)).Run()
 		},
 	}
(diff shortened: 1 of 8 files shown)


pkg/core/stress.go
```diff
@@ -47,6 +47,13 @@ func (s *StressCommand) Validate() error {
 	return nil
 }
 
+func (s *StressCommand) CompleteDefaults() {
+	if s.Workers == 0 {
+		s.Workers = 1
+	}
+
```
Andrewmatilde, line 54: May delete newline here.

Andrewmatilde (approved):
Rest LGTM

pkg/core/stress.go
WangXiangUSTC: done

cmd/attack/stress.go
```diff
@@ -70,11 +71,11 @@ func NewStressMemCommand(dep fx.Option, options *core.StressCommand) *cobra.Comm
 		Short: "continuously stress virtual memory out",
 		Run: func(*cobra.Command, []string) {
 			options.Action = core.StressMemAction
+			options.CompleteDefaults()
 			utils.FxNewAppWithoutLog(dep, fx.Invoke(stressAttackF)).Run()
 		},
 	}
 
-	cmd.Flags().IntVarP(&options.Workers, "workers", "w", 1, "Workers specifies N workers to apply the stressor.")
```
cwen0, line 77: Why delete this flag?

WangXiangUSTC: because every worker will cost some cpu load（almost 100%）, so just use one worker to reduce the cpu cost

cwen0 (approved):
LGTM

PR #79, opened 2021-06-29
jvm attack: support oom on heap and stack

WangXiangUSTC:

Since https://github.com/chaos-mesh/byteman/pull/7 is merged, byteman support out of memory on heap and stack.

This pr add a new config `mem-type` to set `heap` or `stack`, and remove `mem-size`

### DIFF
--- a/cmd/attack/jvm.go
+++ b/cmd/attack/jvm.go
@@ -146,7 +146,7 @@ func NewJVMStressCommand(dep fx.Option, options *core.JVMCommand) *cobra.Command
 	}
 
 	cmd.Flags().IntVarP(&options.CPUCount, "cpu-count", "", 0, "the CPU core number need to use")
-	cmd.Flags().IntVarP(&options.MemorySize, "mem-size", "", 0, "the memory size need to locate, the unit is MB")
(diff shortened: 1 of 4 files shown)


cmd/attack/jvm.go
```diff
@@ -146,7 +146,7 @@ func NewJVMStressCommand(dep fx.Option, options *core.JVMCommand) *cobra.Command
 	}
 
 	cmd.Flags().IntVarP(&options.CPUCount, "cpu-count", "", 0, "the CPU core number need to use")
-	cmd.Flags().IntVarP(&options.MemorySize, "mem-size", "", 0, "the memory size need to locate, the unit is MB")
+	cmd.Flags().StringVarP(&options.MemoryType, "mem-type", "", "", "the memory type need to locate, the value can be 'stack' or 'heap'")
```
fewdan, line 149: Does it only need to specify the type and not the size here?

WangXiangUSTC: Yeah, it is used to support OOM, allocate as much memory as possible until an OOM error occurs, so the size is not required.

PR #84, opened 2021-07-30
support setting uid when injecting fault

WangXiangUSTC:

user can set the uid when injecting fault, it is more convenient for managing the experiment.

### DIFF
--- a/pkg/server/chaosd/attack.go
+++ b/pkg/server/chaosd/attack.go
@@ -51,7 +51,10 @@ func (s *Server) ExecuteAttack(attackType AttackType, options core.AttackConfig,
 		return
 	}
 
-	uid = uuid.New().String()
+	uid = options.GetUID()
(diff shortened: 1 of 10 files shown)


pkg/server/chaosd/attack.go
```diff
@@ -51,7 +51,10 @@ func (s *Server) ExecuteAttack(attackType AttackType, options core.AttackConfig,
 		return
 	}
 
-	uid = uuid.New().String()
+	uid = options.GetUID()
```
fewdan, line 54: Is it because we can directly specify a UID?

fewdan (approved):
rest LGTM

pkg/server/chaosd/attack.go
WangXiangUSTC: Yes

Andrewmatilde (approved):
LGTM

PR #85, opened 2021-08-12
JVM attack: combine install and submit into one

WangXiangUSTC:

combine install and submit into one, to make it easy to create JVM chaos in dashboard


pkg/server/httpserver/server.go
```diff
@@ -201,6 +202,32 @@ func (s *httpServer) createDiskAttack(c *gin.Context) {
...
 
+// @Summary Create jvm attack.
+// @Description Create jvm attack.
+// @Tags attack
+// @Produce json
+// @Param request body core.DiskOption true "Request body"
+// @Success 200 {object} utils.Response
+// @Failure 400 {object} utils.APIError
+// @Failure 500 {object} utils.APIError
+// @Router /api/attack/disk [post]
```
fewdan, line 213: /api/attack/jvm

pkg/server/httpserver/server.go
```diff
@@ -201,6 +202,32 @@ func (s *httpServer) createDiskAttack(c *gin.Context) {
 	c.JSON(http.StatusOK, utils.AttackSuccessResponse(uid))
 }
 
+// @Summary Create jvm attack.
+// @Description Create jvm attack.
+// @Tags attack
+// @Produce json
+// @Param request body core.DiskOption true "Request body"
```
fewdan, line 209: It should not be DiskOption?

fewdan (changes requested):

pkg/server/chaosd/jvm.go
```diff
@@ -60,66 +61,76 @@ const bmInstallCommand = "bminstall.sh -b -Dorg.jboss.byteman.transform.all -Dor
...
-		return err
+		// this error will occured when install agent more than once, and will ignore this error and continue to submit rule
+		errMsg1 := "Agent JAR loaded but agent failed to initialize"
+
+		// these two errors will occured when java version less or euqal to 1.8, and don't know why
+		// but it can install agent success even with this error, so just ignore it now.
+		// TODO: Investigate the cause of these two error
+		errMsg2 := "Provider sun.tools.attach.LinuxAttachProvider not found"
+		errMsg3 := "install java.io.IOException: Non-numeric value found"
+		if !strings.Contains(string(output), errMsg1) && !strings.Contains(string(output), errMsg2) &&
```
Andrewmatilde, line 78: Since errMsgX are all only used once here , I think giving the literal string is enough.

Andrewmatilde (commented):
Rest LGTM

pkg/server/chaosd/jvm.go
WangXiangUSTC: It's convenient to add comments, and avoid lines that are too long

Andrewmatilde: OK.

Andrewmatilde (approved):
LGTM

PR #87, opened 2021-08-24
process attack: support all signal

WangXiangUSTC:

process attack only support signal `SIGKILL`, `SIGTERM`, `SIGSTOP`. In fact, we can support all signal.


Andrewmatilde (approved):
LGTM

Andrewmatilde:
@fewdan PTAL

fewdan (approved):
LGTM

Issue #89, opened 2021-09-02
support bandwidth limit in network attack


WangXiangUSTC:
same with https://chaos-mesh.org/docs/simulate-network-chaos-on-kubernetes/#bandwidth-example

WangXiangUSTC:
close via https://github.com/chaos-mesh/chaosd/pull/91

PR #91, opened 2021-09-15
network attack: support bandwidth limit

WangXiangUSTC:
close issue https://github.com/chaos-mesh/chaosd/issues/89


cmd/attack/network.go
```diff
@@ -191,6 +192,30 @@ func NetworkDNSCommand(dep fx.Option, options *core.NetworkCommand) *cobra.Comma
...
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.NetworkBandwidthAction
+			options.CompleteDefaults()
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(commonNetworkAttackFunc)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Rate, "rate", "r", "", "the speed knob, allows bps, kbps, mbps, gbps, tbps unit. bps means bytes per second")
+	cmd.Flags().Uint32VarP(&options.Limit, "limit", "l", 0, "the number of bytes that can be queued waiting for tokens to become available")
+	cmd.Flags().Uint32VarP(&options.Buffer, "buffer", "b", 0, "the maximum amount of bytes that tokens can be available for instantaneously")
```
YangKeao, line 209: I have a little concern. Some of these parameters are required (like the `rate`, `buffer` and `limit`). Should we give them a better default value (but I cannot tell a reasonable value), or validate the input?

The user may run `chaosd` only with `rate`, and the command will succeed and give back a uid. However, it fails in the `chaos-daemon` (but the error doesn't return out, this bug has been fixed in the latest `chaos-daemon`).

WangXiangUSTC: Yeah, I will add default values for them, and validate the value.

WangXiangUSTC: done, PTAL again

YangKeao (approved):
LGTM

Andrewmatilde (approved):
LGTM

PR #113, opened 2021-12-10
chore: divide http and https server

### LINKED ISSUE #107: HealthCheck API should always be exported by http server
Now, health-check API will return failure if we send http request to chaosd when chaosd using mTLS mode.
https://github.com/chaos-mesh/chaosd/blob/main/pkg/server/httpserver/server.go#L94 
https://github.com/chaos-mesh/chaosd/blob/main/pkg/server/httpserver/auth.go

iguoyr:
Signed-off-by: SiyuChen <ryougi201@gmail.com>

close #107

### DIFF
--- a/pkg/server/httpserver/auth.go
+++ b/pkg/server/httpserver/auth.go
@@ -16,11 +16,12 @@ package httpserver
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"io/ioutil"
+	"net/http"
+
 	"github.com/gin-gonic/gin"
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
-	"io/ioutil"
-	"net/http"
 
 	"github.com/chaos-mesh/chaosd/pkg/server/utils"
 )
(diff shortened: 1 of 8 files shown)


WangXiangUSTC:
@shivanshs9 PTAL

pkg/server/httpserver/auth.go
```diff
@@ -64,66 +70,20 @@ func (s *httpServer) Run(handler func()) (err error) {
 
 		tlsConfig := &tls.Config{
 			ClientCAs:  caCertPool,
-			ClientAuth: tls.RequestClientCert,
+			ClientAuth: tls.RequireAndVerifyClientCert,
```
shivanshs9, line 73: Does using this change the status code for missing certificates and invalid certificates?

shivanshs9 (approved):
LGTM!

pkg/server/httpserver/auth.go
iguoyr: @shivanshs9 Yes, it seems that if tls certs verify failed, it will return with http status code `500` directly.

shivanshs9: @iguoyr Is that fine though? To have a backwards incompatible 500 error for it? Maybe we can keep the previous functions to verify certificates.

iguoyr: Well, I think you're right! Already rollback the code :)

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
> @shivanshs9: Thanks for your review. The bot only counts approvals from reviewers and higher roles in [list](https://prow.tidb.io/tichi/repos/chaos-mesh/chaosd/pulls/113/owners), but you're still welcome to leave your comments.

add @shivanshs9  as reviewer here https://github.com/ti-community-infra/configs/pull/479

STRRL (approved):

PR #126, opened 2022-01-11
Enable fill | write in dir. 

Andrewmatilde:
If path input of disk attack is a dir, we will create a tmp file in the dir & then exec the action.

### DIFF
--- a/pkg/core/disk.go
+++ b/pkg/core/disk.go
@@ -204,7 +204,7 @@ func initPath(opt *DiskOption) (string, error) {
 		if fi.IsDir() {
 			opt.Path, err = utils.CreateTempFile(opt.Path)
 			if err != nil {
-				log.Error(fmt.Sprintf("unexpected err when CreateTempFile in action %s with path %s.", opt.Action, opt.Path))
+				log.Error(fmt.Sprintf("unexpected err : %v , when CreateTempFile in action %s with path %s.", err, opt.Action, opt.Path))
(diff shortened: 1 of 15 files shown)


Andrewmatilde:
PTAL @cwen0 @WangXiangUSTC .

cwen0 (approved):
LGTM

pkg/core/disk.go
```diff
@@ -179,35 +179,40 @@ func initPath(opt *DiskOption) (string, error) {
...
+				}
+				return opt.Path, nil
 			}
+			return "", err
 		}
-		return opt.Path, nil
+		if fi.IsDir() {
+			opt.Path, err = utils.CreateTempFile(opt.Path)
+			if err != nil {
+				log.Error(fmt.Sprintf("unexpected err when CreateTempFile in action %s with path %s.", opt.Action, opt.Path))
```
WangXiangUSTC, line 207: how about adding error information in the log?

WangXiangUSTC (commented):
rest LGTM

pkg/core/disk.go
Andrewmatilde: OK,I agree with you.

Andrewmatilde: Fixed.

WangXiangUSTC:
LGTM

WangXiangUSTC (approved):

PR #130, opened 2022-02-17
Network: support down NIC

FingerLeader:
**Notice:** The parameter `duration` is set to mandatory, because disabling the remote server's NIC is dangerous.

### DIFF
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -57,9 +57,6 @@ type NetworkCommand struct {
 	// only the packet which match the tcp flag can be accepted, others will be dropped.
 	// only set when the IPProtocol is tcp, used for partition.
 	AcceptTCPFlags string `json:"accept-tcp-flags,omitempty"`
-
-	// used for down nic
-	Time string `json:"time,omitempty"`
 }
 
 var _ AttackConfig = &NetworkCommand{}
@@ -212,6 +209,10 @@ func (n *NetworkCommand) validNetworkOccupied() error {
 func (n *NetworkCommand) validNetworkNICDown() error {
 	if len(n.Duration) == 0 {
 		return errors.New("duration is required")
+	}
+
+	if len(n.Device) == 0 {
+		return errors.New("device is required")
 	}
 
 	return nil
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -210,10 +210,10 @@ func (n *NetworkCommand) validNetworkOccupied() error {
 }
 
 func (n *NetworkCommand) validNetworkNICDown() error {
-	if len(n.Time) == 0 {
-		return errors.New("time is required")
+	if len(n.Duration) == 0 {
+		return errors.New("duration is required")
 	}
-	
+
 	return nil
 }
 
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -70,6 +70,7 @@ const (
 	NetworkPartitionAction    = "partition"
 	NetworkBandwidthAction    = "bandwidth"
 	NetworkPortOccupiedAction = "occupied"
+	NetworkNicDownAction      = "down"
 )
 
 func (n *NetworkCommand) Validate() error {
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -57,6 +57,9 @@ type NetworkCommand struct {
 	// only the packet which match the tcp flag can be accepted, others will be dropped.
 	// only set when the IPProtocol is tcp, used for partition.
 	AcceptTCPFlags string `json:"accept-tcp-flags,omitempty"`
+
+	// used for down nic
+	Time string `json:"time,omitempty"`
 }
 
 var _ AttackConfig = &NetworkCommand{}
@@ -70,7 +73,7 @@ const (
 	NetworkPartitionAction    = "partition"
 	NetworkBandwidthAction    = "bandwidth"
 	NetworkPortOccupiedAction = "occupied"
-	NetworkNicDownAction      = "down"
+	NetworkNICDownAction      = "down"
 )
 
 func (n *NetworkCommand) Validate() error {
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -93,6 +93,8 @@ func (n *NetworkCommand) Validate() error {
 		return n.validNetworkOccupied()
 	case NetworkBandwidthAction:
 		return n.validNetworkBandwidth()
+	case NetworkNICDownAction:
+		return n.validNetworkNICDown()
 	default:
 		return errors.Errorf("network action %s not supported", n.Action)
 	}
@@ -204,6 +206,14 @@ func (n *NetworkCommand) validNetworkOccupied() error {
 	if len(n.Port) == 0 {
 		return errors.New("port is required")
 	}
+	return nil
+}
+
+func (n *NetworkCommand) validNetworkNICDown() error {
+	if len(n.Time) == 0 {
+		return errors.New("time is required")
+	}
+	
 	return nil
 }
 
--- a/pkg/server/chaosd/network.go
+++ b/pkg/server/chaosd/network.go
@@ -576,7 +576,7 @@ func (s *Server) getNICIP(attack *core.NetworkCommand) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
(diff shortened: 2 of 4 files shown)


pkg/core/network.go
```diff
@@ -57,6 +57,9 @@ type NetworkCommand struct {
 	// only the packet which match the tcp flag can be accepted, others will be dropped.
 	// only set when the IPProtocol is tcp, used for partition.
 	AcceptTCPFlags string `json:"accept-tcp-flags,omitempty"`
+
+	// used for down nic
+	Time string `json:"time,omitempty"`
```
WangXiangUSTC, line 62: this field is not used

pkg/core/network.go
```diff
@@ -203,6 +209,14 @@ func (n *NetworkCommand) validNetworkOccupied() error {
 	return nil
 }
 
+func (n *NetworkCommand) validNetworkNICDown() error {
```
WangXiangUSTC, line 212: how about validating the `device`?

pkg/server/chaosd/network.go
```diff
@@ -511,3 +530,47 @@ func (s *Server) recoverEtcHosts(attack *core.NetworkCommand, uid string) error
 	}
 	return nil
 }
+
+func (s *Server) recoverNICDown(attack *core.NetworkCommand) error {
+	NICUpCommand := fmt.Sprintf("ifconfig %s %s up", attack.Device, attack.IPAddress)
+
+	recoverCmd := exec.Command("bash", "-c", NICUpCommand)
+	if err := recoverCmd.Start(); err != nil {
```
WangXiangUSTC, line 538: ditto

pkg/server/chaosd/network.go
```diff
@@ -511,3 +530,47 @@ func (s *Server) recoverEtcHosts(attack *core.NetworkCommand, uid string) error
...
+	}
+
+	return nil
+}
+
+func (s *Server) recoverNICDownScheduled(attack *core.NetworkCommand) error {
+	NICUpCommand := fmt.Sprintf("sleep %s && nohup ifconfig %s %s up", attack.Duration, attack.Device, attack.IPAddress)
+
+	recoverCmd := exec.Command("bash", "-c", NICUpCommand)
+	if err := recoverCmd.Start(); err != nil {
```
WangXiangUSTC, line 549: ditto

pkg/server/chaosd/network.go
```diff
@@ -511,3 +530,47 @@ func (s *Server) recoverEtcHosts(attack *core.NetworkCommand, uid string) error
...
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func (s *Server) recoverNICDownScheduled(attack *core.NetworkCommand) error {
```
WangXiangUSTC, line 545: In this way, we need to set the experiment's status to `Destroyed`.

I think it's better to refine the `schedule` framework, and let it recover automatically by the `schedule` framework.

pkg/server/chaosd/network.go
```diff
@@ -85,6 +85,23 @@ func (networkAttack) Attack(options core.AttackConfig, env Environment) (err err
...
+
+	case core.NetworkNICDownAction:
+		if err := env.Chaos.getNICIP(attack); err != nil {
+			return errors.WithStack(err)
+		}
+
+		NICDownCommand := fmt.Sprintf("ifconfig %s down", attack.Device)
+
+		cmd := exec.Command("bash", "-c", NICDownCommand)
+		if err = cmd.Start(); err != nil {
```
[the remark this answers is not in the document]
FingerLeader: But using `cmd.CombinedOutput` or `cmd.Run` will make process stalled when exectuing `sleep $duration`, is it acceptable?

pkg/core/network.go
FingerLeader: I forget it :scream:

pkg/server/chaosd/network.go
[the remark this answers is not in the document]
WangXiangUSTC: I think it's OK. chaosd should't exit before recover it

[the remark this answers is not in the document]
FingerLeader: If chaosd will not exit, I think nohup script is useless. Should I remove it?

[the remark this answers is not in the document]
WangXiangUSTC: Yeah, we can remove the nohup in cmd

pkg/server/chaosd/network.go
WangXiangUSTC: We can refine the `schedule` framework in next pr

pkg/server/chaosd/network.go
```diff
@@ -85,6 +85,25 @@ func (networkAttack) Attack(options core.AttackConfig, env Environment) (err err
...
+		NICDownCommand := fmt.Sprintf("ifconfig %s down", attack.Device)
+
+		cmd := exec.Command("bash", "-c", NICDownCommand)
+		stdout, err := cmd.CombinedOutput()
+		if err != nil {
+			log.Error(cmd.String()+string(stdout), zap.Error(err))
+			return errors.WithStack(err)
+		}
+
+		if attack.Duration != "-1" {
```
WangXiangUSTC, line 103: is this should be `len(attack.Duration) != 0`?

WangXiangUSTC (commented):
rest LGTM

pkg/server/chaosd/network.go
FingerLeader: `len(attack.Duration)` has been checked before this function, `-1` means attack will run continuously. It will be marked in doc.

WangXiangUSTC (approved):
LGTM

pkg/server/chaosd/network.go
cwen0: Why not check this field before executing this attack?

pkg/server/chaosd/network.go
```diff
@@ -511,3 +532,51 @@ func (s *Server) recoverEtcHosts(attack *core.NetworkCommand, uid string) error
...
 	return nil
 }
+
+func (s *Server) recoverNICDown(attack *core.NetworkCommand) error {
+	NICUpCommand := fmt.Sprintf("ifconfig %s %s up", attack.Device, attack.IPAddress)
+
+	recoverCmd := exec.Command("bash", "-c", NICUpCommand)
+	stdout, err := recoverCmd.CombinedOutput()
+	if err != nil {
+		log.Error(recoverCmd.String()+string(stdout), zap.Error(err))
```
cwen0, line 542: We can delete this log. refer to https://github.com/chaos-mesh/rfcs/blob/main/text/2021-12-09-logging.md#relations-between-error-and-logging

pkg/server/chaosd/network.go
FingerLeader: This field has been checked for necessity before executing, `if len(n.Duration) == 0 {
		return errors.New("duration is required")
	}`, the judgement here is only to decide whether to run continuously.

pkg/server/chaosd/network.go
FingerLeader: I removed these `log.Error`, PTAL : )

pkg/server/chaosd/network.go
FingerLeader: Actually, this `-1` option is added on purpose, because I think some user may need this option to let the experiment run continuously, especially those who run NIC down locally rather than remotely. And `attack.Duration` is hardly ever set to negative numbers, I think `-1` is enough to be a warning sign (maybe `-2333` or `-32768` is better).

pkg/server/chaosd/network.go
```diff
@@ -511,3 +531,49 @@ func (s *Server) recoverEtcHosts(attack *core.NetworkCommand, uid string) error
...
+	if err = cmd.Start(); err != nil {
+		return errors.WithStack(err)
+	}
+
+	stdoutBytes := make([]byte, 1024)
+	_, err = stdout.Read(stdoutBytes)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	attack.IPAddress = strings.TrimRight(string(stdoutBytes), "\n\x00")
```
cwen0, line 576: Add some comments to explain why need to trim `\x00`

cwen0 (approved):
LGTM

Issue #127, opened 2022-01-17
network partition: direction support "both"


WangXiangUSTC:
`direction` can only be set to `from` or `to`, need to support `both`, to support bidirectional network partition.

PR #136, opened 2022-03-01
support "both" direction for network partition

WangXiangUSTC:
Now only support direction "from" and "to" for partition, users need to run chaosd command twice if they want to partition both "from" and "to". 

This pr close issue #127


pkg/core/network.go
```diff
@@ -469,12 +469,37 @@ func (n *NetworkCommand) PartitionChain(ipset string) ([]*pb.Chain, error) {
...
+	}
+
+	if n.Direction == "from" || n.Direction == "both" {
+		fromChains, err = n.getPartitionChain(ipset, "from")
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+	}
+
+	chains = append(chains, toChains...)
```
Andrewmatilde, line 490: It seems an empty Chain may appended here.

pkg/core/network.go
```diff
@@ -497,6 +522,7 @@ func (n *NetworkCommand) PartitionChain(ipset string) ([]*pb.Chain, error) {
 		Name:      fmt.Sprintf("%s/1", directionStr),
 		Ipsets:    []string{ipset},
 		Direction: directionChain,
+		Protocol:  n.IPProtocol,
```
Andrewmatilde, line 525: I think protocol maybe not meaningful here , since if one of input & output chain drop , (tcp,udp,icmp) will all broken.

Andrewmatilde (commented):
Rest LGTM

pkg/core/network.go
WangXiangUSTC: checked the `Direction` in function `validNetworkPartition`, it must be one of `from`, `to` and `both`

pkg/core/network.go
WangXiangUSTC: `iptables` can drop network package by specified protocol, the command looks like `iptables -w -A OUTPUT/1 -m set --match-set chaos-962d44ca-a0dc-4f dst -j DROP -w 5 --protocol tcp`, so I think the `protocol` here is meaningful

Andrewmatilde: Sorry , I misunderstand it. I agree with you.

pkg/core/network.go
Andrewmatilde: OK, I agree with you.

Andrewmatilde (approved):
LGTM

WangXiangUSTC:
@cwen0 PTAL again

pkg/core/network.go
```diff
@@ -469,12 +469,37 @@ func (n *NetworkCommand) PartitionChain(ipset string) ([]*pb.Chain, error) {
...
 	}
 
+	chains := make([]*pb.Chain, 0, 2)
+	var toChains, fromChains []*pb.Chain
+	var err error
+
+	if n.Direction == "to" || n.Direction == "both" {
+		toChains, err = n.getPartitionChain(ipset, "to")
+		if err != nil {
+			return nil, errors.WithStack(err)
```
cwen0, line 479: If the error is returned by function inside the Chaos Mesh, we could assume this error is already equipped with a stack backtrace, so there is no need to call `errors.WithStack`.

pkg/core/network.go
```diff
@@ -469,12 +469,37 @@ func (n *NetworkCommand) PartitionChain(ipset string) ([]*pb.Chain, error) {
...
+		}
+	}
+
+	if n.Direction == "from" || n.Direction == "both" {
+		fromChains, err = n.getPartitionChain(ipset, "from")
+		if err != nil {
+			return nil, errors.WithStack(err)
```
cwen0, line 486: ditto

pkg/core/network.go
WangXiangUSTC: The function `getPartitionChain` only returns an error when `direction` is neither `to` nor `from`,   the error is not from Chaos Mesh

WangXiangUSTC:
@cwen0 PTAL again

pkg/core/network.go
cwen0: The error of `getPartitionChain` was generated by `errors.New`, which has equipped with a stack backtrace. So we don't need to use 'errors.WithStack` here.

cwen0 (approved):
LGTM

PR #145, opened 2022-03-24
ignore process not found while recovering the stress

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

If the process is not found, we can assume that it's already been recovered (or killed by other processes, like oom killer).

### DIFF
--- a/pkg/server/chaosd/stress.go
+++ b/pkg/server/chaosd/stress.go
@@ -15,7 +15,7 @@ package chaosd
 
 import (
 	"errors"
-	"os"
+	"io/fs"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


pkg/server/chaosd/stress.go
```diff
@@ -91,6 +92,10 @@ func (stressAttack) Recover(exp core.Experiment, _ Environment) error {
 	attack := config.(*core.StressCommand)
 	proc, err := process.NewProcess(attack.StressngPid)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
iguoyr, line 95: According to [this comment](https://github.com/golang/go/blob/master/src/os/error.go#L92), it suggests us to use `fs.ErrNotExist` instead of `os.ErrNotExist`.

iguoyr (commented):
Rest LGTM

pkg/server/chaosd/stress.go
YangKeao: Good! I will use `os.IsNotExist`

Andrewmatilde (approved):
LGTM

WangXiangUSTC (approved):
LGTM

WangXiangUSTC (approved):

WangXiangUSTC (approved):

iguoyr (approved):
LGTM

Issue #133, opened 2022-02-21
inject fault into Redis


WangXiangUSTC:
include:
- switch primary and secondary
- batch delete
- cache hotspot
- sentinel is unavaliable
- sentinel restart

FingerLeader:
In pr #152 , implemented `sentinel-stop` and `sentinel-restart` by using functions provided in pkg/go-redis

PR #152, opened 2022-04-07
redisChaos: support sentinel stop/restart

FingerLeader:


cwen0:
@FingerLeader Can you fix the CI first?

pkg/server/chaosd/redis.go
```diff
@@ -0,0 +1,107 @@
...
+		result, err := sentinelCli.FlushConfig(sentinelCli.Context()).Result()
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		if result != STATUSOK {
+			return errors.WithStack(errors.Errorf("redis command status is %s", result))
+		}
+	}
+
+	// If cli.Shutdown() runs successfully, the result will be nil and the err will be "connection refused"
```
WangXiangUSTC, line 88: This looks strange 🤔

pkg/server/chaosd/redis.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return nil
+}
+
+func (s *Server) recoverSentinelStop(attack *core.RedisCommand) error {
+	if attack.Conf == "" {
+		return errors.WithStack(errors.Errorf("redis config is not present"))
+	}
+	recoverCmd := exec.Command("redis-server", attack.Conf, "--sentinel")
```
WangXiangUSTC, line 101: We need to make sure the `redis-server` is in `PATH`. I think we can provide a `recover-cmd` just like the `process` attack.

pkg/server/chaosd/redis.go
```diff
@@ -0,0 +1,107 @@
...
+		return errors.WithStack(err)
+	}
+
+	switch attack.Action {
+	case core.RedisSentinelRestartAction:
+		err := env.Chaos.shutdownSentinelServer(attack, cli)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		return env.Chaos.recoverSentinelStop(attack)
```
WangXiangUSTC, line 51: seems that `restart` only does this one more thing than `stop`,  I think we can only remain one action.

pkg/server/chaosd/redis.go
FingerLeader: Yep, the return value is strange, but I need something to make sure the server is shut down

WangXiangUSTC: Seems that there is no other better function to use. This LGTM

pkg/server/chaosd/redis.go
FingerLeader: Do you mean remove `sentinel stop`? I'm not sure that's a good decision, because `stop` and 'restart` are two separate functions in cloud chaos.

WangXiangUSTC: It's ok to keep the `sentinel stop`

pkg/server/chaosd/redis.go
FingerLeader: Added a flag to set the path of `redis-server`

FingerLeader:
PTAL /cc @cwen0 @WangXiangUSTC

WangXiangUSTC (approved):
LGTM

cmd/attack/redis.go
```diff
@@ -0,0 +1,98 @@
...
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.RedisSentinelStopAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(redisAttackF)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
+	cmd.Flags().StringVarP(&options.Conf, "conf", "c", "", "The config path of Redis server")
+	cmd.Flags().BoolVarP(&options.FlushConfig, "flush-config", "", true, " Force Sentinel to rewrite its configuration on disk")
```
WangXiangUSTC, line 83:
```suggestion
	cmd.Flags().BoolVarP(&options.FlushConfig, "flush-config", "", true, "Force Sentinel to rewrite its configuration on disk")
```

WangXiangUSTC:
It's better to add some description for this pr, and link the issue https://github.com/chaos-mesh/chaosd/issues/133

FingerLeader:
> It's better to add some description for this pr, and link the issue #133

Done it

cwen0 (approved):
LGTM

PR #154, opened 2022-04-14
implement kafka chaos

Hexilee:
Add three subcommands for kafka attack:
#### Usage
```bash
Flood kafka cluster with messages

Usage:
  chaosd attack kafka flood [options] [flags]

Flags:
  -h, --help              help for flood
  -H, --host string       the host of kafka leader (default "localhost")
  -p, --password string   the password of kafka client
  -P, --port uint16       the port of kafka leader (default 9092)
  -r, --rps uint          the request per second for each worker (default 128)
  -s, --size uint         the size of each message (default 1024)
  -t, --threads uint      the numbers of worker threads (default 100)
  -u, --username string   the username of kafka client

Global Flags:
  -T, --topic string       the topic to attack
```

#### Example

```bash
> bin/chaosd attack kafka flood -T test -r 200 -t 12
```

#### Usage
```bash
Fill kafka cluster with messages

Usage:
  chaosd attack kafka fill [options] [flags]

Flags:
  -h, --help              help for fill
  -H, --host string       the host of kafka leader (default "localhost")
  -m, --max-bytes uint    the max bytes to fill (default 17179869184)
  -p, --password string   the password of kafka client
  -P, --port uint16       the port of kafka leader (default 9092)
  -s, --size uint         the size of each message (default 4096)
  -u, --username string   the username of kafka client

Global Flags:
  -T, --topic string       the topic to attack
```

#### Example

```bash
> bin/chaosd attack kafka fill -T test -m 1000_000_000
```

#### Usage
```bash
Make kafka cluster non-writable/non-readable

Usage:
  chaosd attack kafka io [options] [flags]

Flags:
  -c, --config string   the path of server config (default "/etc/kafka/server.properties")
  -h, --help            help for io
  -r, --non-readable    make kafka cluster non-readable
  -w, --non-writable    make kafka cluster non-writable

Global Flags:
      --partition int      the partition to attack
  -T, --topic string       the topic to attack
```
#### Example

```bash
bin/chaosd attack kafka io -T test -rw
```


Hexilee:
PTAL @WangXiangUSTC

cmd/attack/kafka.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2020 Chaos Mesh Authors.
```
cwen0, line 1: s/2020/2022/g

Hexilee: Fixed

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
+// Copyright 2020 Chaos Mesh Authors.
```
cwen0, line 1: ditto

pkg/core/kafka.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2021 Chaos Mesh Authors.
```
cwen0, line 1: s/2021/2022/g

cmd/attack/kafka.go
```diff
@@ -0,0 +1,125 @@
...
+	)
+
+	cmd := &cobra.Command{
+		Use:   "kafka <subcommand>",
+		Short: "Kafka attack related commands",
+	}
+
+	cmd.PersistentFlags().StringVarP(&options.Topic, "topic", "T", "", "the topic to attack")
+	cmd.PersistentFlags().IntVarP(&options.Partition, "partition", "", 0, "the partition to attack")
+	cmd.PersistentFlags().BoolVarP(&options.NoSilent, "no-silent", "", false, "print more information")
```
WangXiangUSTC, line 45: Why need this config? Seems that can use log-level to replace.

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
...
+	attack := options.(*core.KafkaCommand)
+	switch attack.Action {
+	case core.KafkaFillAction:
+		return attackKafkaFill(attack)
+	case core.KafkaFloodAction:
+		return attackKafkaFlood(context.TODO(), attack)
+	case core.KafkaIOAction:
+		return attackKafkaIO(attack)
+	default:
+		return nil
```
WangXiangUSTC, line 51: it might be better to return error

cmd/attack/kafka.go
```diff
@@ -0,0 +1,125 @@
...
+		}),
+	)
+
+	cmd := &cobra.Command{
+		Use:   "kafka <subcommand>",
+		Short: "Kafka attack related commands",
+	}
+
+	cmd.PersistentFlags().StringVarP(&options.Topic, "topic", "T", "", "the topic to attack")
+	cmd.PersistentFlags().IntVarP(&options.Partition, "partition", "", 0, "the partition to attack")
```
WangXiangUSTC, line 44: if the partition is 0(default), chaosd will only write messages to this partition, is it right?

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
...
+	conn, err = dialer.DialLeader(context.Background(), "tcp", endpoint, attack.Topic, attack.Partition)
+	if err != nil {
+		return nil, perr.Wrapf(err, "dial kafka leader %s", endpoint)
+	}
+	return conn, nil
+}
+
+func attackKafkaFill(attack *core.KafkaCommand) (err error) {
+	// TODO: make it configurable
+	const messagePerRequest = 128
```
WangXiangUSTC, line 87: seems that it is already configurable

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	mode := dir.Mode()
+	if attack.NonReadable {
+		mode &= ^os.FileMode(0444)
+	}
+	if attack.NonWritable {
+		mode &= ^os.FileMode(0200)
+	}
+	if attack.NoSilent {
+		log.Info(fmt.Sprintf("change permission of %s to %s", attack.PartitionDir, mode))
```
WangXiangUSTC, line 198: can you log.Infof

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
...
+
+		mode := f.Mode()
+		if attack.NonReadable {
+			mode &= ^os.FileMode(0444)
+		}
+		if attack.NonWritable {
+			mode &= ^os.FileMode(0200)
+		}
+		if attack.NoSilent {
+			log.Info(fmt.Sprintf("change permission of %s to %s", filePath, mode))
```
WangXiangUSTC, line 227: ditto

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,303 @@
...
+							if attack.NoSilent {
+								logger.Error("write message", zap.Error(err))
+							}
+							continue
+						}
+						succeeded++
+					}
+					logger.Info(fmt.Sprintf("succeeded: %d, failed: %d", succeeded, failed))
+					if time.Now().Sub(start) < time.Second {
+						time.Sleep(start.Add(time.Second).Sub(time.Now()))
```
WangXiangUSTC, line 165: why need to sleep?

pkg/core/kafka.go
```diff
@@ -0,0 +1,137 @@
...
+	// options for io attack
+	ConfigFile  string
+	NonReadable bool
+	NonWritable bool
+
+	// recover data for io attack
+	PartitionDir string
+}
+
+func (c *KafkaCommand) Validate() error {
```
cwen0, line 64: Can we split this function into multiple functions?

pkg/server/chaosd/kafka.go
Hexilee: Fixed

pkg/core/kafka.go
Hexilee: Fixed

cmd/attack/kafka.go
Hexilee: Removed

pkg/server/chaosd/kafka.go
Hexilee: Fixed

cmd/attack/kafka.go
Hexilee: Yes

pkg/server/chaosd/kafka.go
Hexilee: Fixed

pkg/server/chaosd/kafka.go
Hexilee: Fixed

pkg/server/chaosd/kafka.go
Hexilee: No, it's non-configurable. The `messagePerRequest` means the number of messages in a single request.

pkg/server/chaosd/kafka.go
Hexilee: To control the request-per-second. If `time.Now().Sub(start) < time.Second`, the target number of requests should be complete, so we should sleep for the next second.

pkg/core/kafka.go
Hexilee: Split.

cmd/attack/kafka.go
WangXiangUSTC: Writing messages to only one partition will reduce the performance. How about using `-1` as the default value, and when the partition is -1,  write messages to the topic and without specify a partition

Hexilee: Oops, I do not find relevant logic in the docs or source code of [` LookupPartition`](https://github.com/segmentio/kafka-go/blob/v0.4.31/dialer.go#L162) and [`DialLeader`](https://github.com/segmentio/kafka-go/blob/v0.4.31/dialer.go#L132)

Hexilee: Done

cmd/attack/kafka.go
```diff
@@ -0,0 +1,125 @@
...
+		Short:             "Fill kafka cluster with messages",
+		ValidArgsFunction: cobra.NoFileCompletions,
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.KafkaFillAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(kafkaCommandFunc)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Host, "host", "H", "localhost", "the host of kafka leader")
+	cmd.Flags().Uint16VarP(&options.Port, "port", "P", 9092, "the port of kafka leader")
```
WangXiangUSTC, line 67: it seems that we can combine these two configs into one, like `address`

WangXiangUSTC: Another question, is whether it must be set to the leader's address?

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,381 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-errChan:
+			return err
+		case n := <-writeChan:
+			bytes += uint64(n)
+			newWritten := humanize.Bytes(bytes)
+			if newWritten != written {
```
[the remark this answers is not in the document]
Hexilee: To ignore printing duplicated information.

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		partitions = append(partitions, par.ID)
+	}
+	return partitions, nil
+}
+
+func attackKafkaFill(ctx context.Context, attack *core.KafkaCommand) (err error) {
+	// TODO: make it configurable
+	const messagePerRequest = 128
+	conn, err := dial(attack)
```
cwen0, line 138: Do we need to use a conn pool here? I see that only one conn is actually created, but the following is to use multiple goroutines to send messages.

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,381 @@
...
+
+func recoverKafkaIO(attack *core.KafkaCommand, env Environment) (err error) {
+	for _, dirPath := range attack.PartitionDirs {
+		dir, err := os.Stat(dirPath)
+		if err != nil {
+			return perr.Wrapf(err, "stat partition dir %s", dirPath)
+		}
+		mode := dir.Mode()
+		if attack.NonReadable {
+			mode |= os.FileMode(0444)
```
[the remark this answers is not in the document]
Hexilee: I'm going to store the origin permission of all files.

cmd/attack/kafka.go
Hexilee: > it seems that we can combine these two configs into one, like address
I prefer to not combine them, as many CLI tools like `mysql-client`, `ssh`, etc.

> Another question, is whether it must be set to the leader's address?
It seems unnecessary, I will change the description.

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	dirs := make([]string, 0, len(partitions))
+
+	for _, partition := range partitions {
+		dirName := fmt.Sprintf("%s-%d", attack.Topic, partition)
+		for _, dir := range logDirs {
+			entries, err := os.ReadDir(strings.TrimSpace(dir))
+			if err != nil {
+				log.Debug("read dir", zap.Error(err))
+				continue
```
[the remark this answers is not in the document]
Hexilee: Yes.

pkg/server/chaosd/kafka.go
cwen0: Oh, my fault, but I found this `conn` was not used in the following logic.

Hexilee: Removed

pkg/server/chaosd/kafka.go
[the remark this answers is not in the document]
Hexilee: Returned

pkg/server/chaosd/kafka.go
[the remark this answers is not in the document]
Hexilee: Now, we recover the origin mode by storing the origin permission of all files (Include unit test).

cmd/attack/kafka.go
Hexilee: The description is changed.

pkg/server/chaosd/kafka.go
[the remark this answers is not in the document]
WangXiangUSTC: Every time get `n` from channel `writeChan` means had already written some message to Kafka, I think it's not necessary to add this judgment, print the log below even if chaosd write same size message with last time

pkg/server/chaosd/kafka.go
```diff
@@ -0,0 +1,353 @@
...
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					start := time.Now()
+					succeeded := 0
+					failed := 0
+					counter := uint64(0)
+					for time.Now().Sub(start) < time.Second && counter < attack.RequestPerSecond {
```
WangXiangUSTC, line 222: In my opinion, we can remove the config `RequestPerSecond`, chaosd can write messages as fast as possible, and then we can remove this complex logic below.

WangXiangUSTC (commented):
rest LGTM

pkg/server/chaosd/kafka.go
Hexilee: > Every time get n from channel writeChan means had already written some message to Kafka, I think it's not necessary to add this judgment, print the log below even if chaosd write same size message with last time

Because we use the humanize to format bytes, there may be the same `written` in many lines when the written is too large.

WangXiangUSTC (approved):
LGTM

Hexilee:
@cwen0 All comments are resolved, PTAL again!

cwen0 (approved):
LGTM

PR #155, opened 2022-04-18
bump gopsutil to 3.21.11

FingerLeader:
Because the original version of gopsutil does not support `TARGET_OS_MAC`, I think we can bump it to the same version(v3.21.11) as chaos mesh.


WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
@Andrewmatilde PTAL

Andrewmatilde (approved):
LGTM

PR #157, opened 2022-04-27
JVM attack: add json label for MySQL config

WangXiangUSTC:

1. add json label for MySQL config in JVM attack
2. minor fix on the description of duration


cwen0 (approved):
LGTM

Issue #25, opened 2021-03-04
support kill vm


WangXiangUSTC:
chaosd need to support kill vm, can refer to https://github.com/pingcap/tidb-operator/blob/845a716d41ad1e1ae2579048e50f8a9516e975ab/tests/pkg/fault-trigger/manager/vm_virsh.go#L45

PR #158, opened 2022-05-05
support vm attack

WangXiangUSTC:

close issue #25 

usage: 

```shell
./bin/chaosd attack vm -h
vm attack by using virsh

Usage:
  chaosd attack vm attack [flags]

Flags:
  -h, --help             help for vm
  -v, --vm-name string   The name of the vm to be destoryed

Global Flags:

```


Andrewmatilde:
In my opinion,  codes for downloading virsh should be added in Makefile.

Andrewmatilde:
Rest LGTM

WangXiangUSTC:
> In my opinion, codes for downloading virsh should be added in Makefile.

Users need to download `virsh` by `yum` or `apt`, depending on the Linux distribution. I think we can add the download command in the document

Andrewmatilde (approved):
OK, I agree with you , LGTM

cwen0 (approved):
LGTM

Andrewmatilde (approved):
LGTM

PR #159, opened 2022-05-06
Support Redis cache penetration

FingerLeader:


pkg/server/chaosd/redis.go
```diff
@@ -52,6 +52,16 @@ func (redisAttack) Attack(options core.AttackConfig, env Environment) error {
 
 	case core.RedisSentinelStopAction:
 		return env.Chaos.shutdownSentinelServer(attack, cli)
+
+	case core.RedisCachePenetrationAction:
+		pipe := cli.Pipeline()
+		for i := 0; i < attack.RequestNum; i++ {
+			pipe.Get(cli.Context(), "CHAOS_MESH_nqE3BWm7khHv")
```
fewdan, line 59: Just curious, what is this "CHAOS_MESH_nqE3BWm7khHv"?

FingerLeader: It's a meaningless code to make sure this key doesn't have a paired value.

fewdan: Oh I see, thanks.

pkg/server/chaosd/redis.go
```diff
@@ -52,6 +52,16 @@ func (redisAttack) Attack(options core.AttackConfig, env Environment) error {
...
+		}
+		_, err := pipe.Exec(cli.Context())
+		if err.Error() != "redis: nil" {
```
WangXiangUSTC, line 62: how about using `redis.Nil`?
https://redis.uptrace.dev/guide/go-redis.html#redis-nil

WangXiangUSTC (commented):
rest LGTM

cmd/attack/redis.go
```diff
@@ -85,6 +86,23 @@ func NewRedisSentinelStopCommand(dep fx.Option, options *core.RedisCommand) *cob
...
+		Short: "penetrate cache",
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.RedisCachePenetrationAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(redisAttackF)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
+	cmd.Flags().IntVarP(&options.RequestNum, "requestnum", "", 0, "The number of requests")
```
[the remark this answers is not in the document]
FingerLeader: Updated to `request-num`

pkg/server/chaosd/redis.go
FingerLeader: Updated to `redis.Nil`

FingerLeader:
PTAL /cc @cwen0

WangXiangUSTC (approved):
LGTM

pkg/server/chaosd/redis.go
cwen0: Why not use a random string?

cmd/attack/redis.go
```diff
@@ -85,6 +86,23 @@ func NewRedisSentinelStopCommand(dep fx.Option, options *core.RedisCommand) *cob
...
+		Short: "penetrate cache",
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.RedisCachePenetrationAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(redisAttackF)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
+	cmd.Flags().IntVarP(&options.RequestNum, "request-num", "", 0, "The number of requests")
```
cwen0, line 101: I think `0` is not a good default value.

FingerLeader: Because I think generating a random code is an unnecessary operation to simulate cache penetration. On the other hand, generating random code will bring additional time consumption which may make the loop runs slowly(especially when the request-num is very big). But I do accept the idea of using random codes, it's not a big deal.

pkg/server/chaosd/redis.go
FingerLeader: Because I think generating a random code is an unnecessary operation to simulate cache penetration. On the other hand, generating random code will bring additional time consumption which may make the loop runs slowly(especially when the request-num is very big). But I do accept the idea of using random codes, it's not a big deal.

cwen0:
@FingerLeader Please fix ci first. Thanks!

cwen0 (approved):
LGTM

PR #163, opened 2022-05-12
redisChaos: add cache limit

FingerLeader:


cmd/attack/redis.go
```diff
@@ -85,6 +86,23 @@ func NewRedisSentinelStopCommand(dep fx.Option, options *core.RedisCommand) *cob
...
+		Short: "set maxmemory of Redis",
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.RedisCacheLimitAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(redisAttackF)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
+	cmd.Flags().StringVarP(&options.CacheSize, "size", "s", "0", "The size of cache")
```
WangXiangUSTC, line 101: what will happen if the cache size is 0? Do we need to check the value in validation?

pkg/server/chaosd/redis.go
```diff
@@ -52,6 +53,22 @@ func (redisAttack) Attack(options core.AttackConfig, env Environment) error {
...
 	case core.RedisSentinelStopAction:
 		return env.Chaos.shutdownSentinelServer(attack, cli)
+
+	case core.RedisCacheLimitAction:
+		cacheSize, err := cli.ConfigGet(cli.Context(), "maxmemory").Result()
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		// Get the value of maxmemory
+		attack.OriginCacheSize = fmt.Sprint(cacheSize[1])
```
WangXiangUSTC, line 63: seems that `cacheSize` is an array,  what is the `cacheSize[0]`?

cmd/attack/redis.go
FingerLeader: If user doesn't set `maxmemory` in redis config, it will be `0`, it is a default value. And redis will help us to validate this value, I don't think we need to do extra validation.

pkg/server/chaosd/redis.go
FingerLeader: Because `redis.SetConfigSet().Result()` returns an interface list like `[maxmemory 100MB]`, the second one is the value what we needed.

WangXiangUSTC: OK, it's better to add a comment for it, add the example of the cacheSize

cmd/attack/redis.go
WangXiangUSTC: OK

pkg/server/chaosd/redis.go
FingerLeader: Updated

WangXiangUSTC (approved):
LGTM

pkg/server/chaosd/redis.go
```diff
@@ -73,9 +108,23 @@ func (redisAttack) Recover(exp core.Experiment, env Environment) error {
 	}
 	attack := config.(*core.RedisCommand)
 
+	cli := redis.NewClient(&redis.Options{
```
cwen0, line 111: If the `action` is `core.RedisSentinelStopAction`, this logic is useless. So we should add it to option `core.RedisCacheLimitAction:` logic.

FingerLeader: Updated!

cwen0:
@FingerLeader CI broken! Please fix it!

cwen0 (approved):
LGTM

Issue #169, opened 2022-05-19
support user-defined attack


WangXiangUSTC:
Users set the attack command and recover command, chaosd execute these command when attack or recover

PR #170, opened 2022-05-20
support user defined attack

WangXiangUSTC:
Close issue #169 

Users set the `attack-cmd` and `recover-cmd`, execute `attack-cmd` when attack, execute `recover-cmd` when recover.

```bash
chaosd attack user-defined -h
user defined attack related commands

Usage:
  chaosd attack user-defined <subcommand> [flags]

Flags:
  -a, --attack-cmd string    the command to be executed when attack
  -h, --help                 help for user-defined
  -r, --recover-cmd string   the command to be executed when recover

Global Flags:
```


Andrewmatilde (approved):
LGTM

test/integration_test/user_defined/run.sh
```diff
@@ -0,0 +1,38 @@
...
+
+cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+cd $cur
+
+bin_path=../../../bin
+
+# test user defined attack
+${bin_path}/chaosd attack user-defined --attack-cmd "touch /tmp/chaos" --recover-cmd "rm /tmp/chaos" --uid 1234567890
+
+if [ ! -f /tmp/chaos ]; then
```
fewdan, line 26: If other progress just created a directory called "/tmp/chaos", then "touch /tmp/chaos" will not report an error, but [ ! -f /tmp/chaos ] will get true, so is it possible to use '- e' would be more appropriate?

test/integration_test/user_defined/run.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 Chaos Mesh Authors.
```
fewdan, line 3: Change to "2022"?

fewdan (approved):
rest LGTM

cwen0 (approved):
LGTM

test/integration_test/user_defined/run.sh
WangXiangUSTC: done

PR #175, opened 2022-05-30
RadisChaos: Add Cache Expiration

FingerLeader:

### DIFF
--- a/cmd/attack/redis.go
+++ b/cmd/attack/redis.go
@@ -60,7 +60,7 @@ func NewRedisSentinelRestartCommand(dep fx.Option, options *core.RedisCommand) *
 		},
 	}
 
-	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "The address of redis server")
 	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
 	cmd.Flags().StringVarP(&options.Conf, "conf", "c", "", "The config of Redis server")
 	cmd.Flags().BoolVarP(&options.FlushConfig, "flush-config", "", true, " Force Sentinel to rewrite its configuration on disk")
@@ -79,7 +79,7 @@ func NewRedisSentinelStopCommand(dep fx.Option, options *core.RedisCommand) *cob
 		},
 	}
 
-	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "The address of redis server")
 	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
 	cmd.Flags().StringVarP(&options.Conf, "conf", "c", "", "The config path of Redis server")
 	cmd.Flags().BoolVarP(&options.FlushConfig, "flush-config", "", true, "Force Sentinel to rewrite its configuration on disk")
@@ -97,7 +97,7 @@ func NewRedisCachePenetrationCommand(dep fx.Option, options *core.RedisCommand) 
 		},
 	}
 
-	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "")
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "The address of redis server")
 	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
 	cmd.Flags().IntVarP(&options.RequestNum, "request-num", "", 0, "The number of requests")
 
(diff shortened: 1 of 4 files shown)


pkg/core/redis.go
```diff
@@ -36,6 +37,9 @@ type RedisCommand struct {
 	FlushConfig bool   `json:"flushConfig,omitempty"`
 	RedisPath   string `json:"redisPath,omitempty"`
 	RequestNum  int    `json:"requestNum,omitempty"`
+	Key         string `json:"key,omitempty"`
```
WangXiangUSTC, line 40: seems that the `key` is useless

pkg/server/chaosd/redis.go
```diff
@@ -121,3 +125,69 @@ func (s *Server) recoverSentinelStop(attack *core.RedisCommand) error {
...
+	}
+
+	expiration, err := time.ParseDuration(attack.Expiration)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if attack.Option == "NX" {
+		for _, key := range allKeys {
+			result, err := cli.ExpireNX(cli.Context(), key, expiration).Result()
```
WangXiangUSTC, line 143: How about defining a variable `ExpireFunc`, and setting it to a specified function depending on the `attack.Option`.  For example, if the option is "NX", set it to `cli.ExpireNX`. In this way, we can remove some redundant codes.

pkg/core/redis.go
FingerLeader: I should add a "`key` is `$1`" judgement, only when the `key` is `$1` to delete all the keys, otherwire only delete the `key`.

pkg/server/chaosd/redis.go
```diff
@@ -121,3 +125,69 @@ func (s *Server) recoverSentinelStop(attack *core.RedisCommand) error {
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	expiration, err := time.ParseDuration(attack.Expiration)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if attack.Option == "NX" {
```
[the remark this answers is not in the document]
FingerLeader: I'm not sure whether it is necessary to explain `option` in comments, because it is the same as redis offical documents.

FingerLeader:
I find `option` can not work well, I will figure it out as soon as possible.

FingerLeader:
> I find `option` can not work well, I will figure it out as soon as possible.

It works now

pkg/core/redis.go
WangXiangUSTC: OK, I think it's better to update the description of `key` in flag

pkg/server/chaosd/redis.go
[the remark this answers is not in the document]
WangXiangUSTC: just a link of Redis official document is OK

WangXiangUSTC (commented):
Please fix the ci, rest LGTM

WangXiangUSTC (approved):
LGTM

cmd/attack/redis.go
```diff
@@ -116,14 +117,33 @@ func NewRedisCacheLimitCommand(dep fx.Option, options *core.RedisCommand) *cobra
...
+		Run: func(*cobra.Command, []string) {
+			options.Action = core.RedisCacheExpirationAction
+			utils.FxNewAppWithoutLog(dep, fx.Invoke(redisAttackF)).Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.Addr, "addr", "a", "", "The address of redis server")
+	cmd.Flags().StringVarP(&options.Password, "password", "p", "", "The password of server")
+	cmd.Flags().StringVarP(&options.Key, "key", "k", "", "The key to be set a expiration, default expire all keys")
+	cmd.Flags().StringVarP(&options.Expiration, "expiration", "", "0", "The expiration of the key")
```
cwen0, line 141: Does this flag represent expiration time?  Can we add an example of this flag?

pkg/core/redis.go
```diff
@@ -60,6 +64,10 @@ func (r *RedisCommand) Validate() error {
 		if r.CacheSize != "0" && r.Percent != "" {
 			return errors.New("only one of cachesize and percent can be set")
 		}
+	case RedisCacheExpirationAction:
+		if r.Option != "" && r.Option != "XX" && r.Option != "NX" && r.Option != "GT" && r.Option != "LT" {
```
cwen0, line 68: This check is ugly. We can add use a map or an array to improve it.

pkg/server/chaosd/redis.go
```diff
@@ -170,3 +174,52 @@ func (s *Server) recoverSentinelStop(attack *core.RedisCommand) error {
...
+		if !result {
+			return errors.WithStack(errors.Errorf("expire failed"))
+		}
+	}
+
+	return nil
+}
+
+func ExpireFunc(cli *redis.Client, key string, expiration time.Duration, option string) *redis.BoolCmd {
+	if option == "NX" {
```
cwen0, line 214: I prefer to use `switch`  instead of `if else`

cwen0:
@FingerLeader PTAL

cmd/attack/redis.go
FingerLeader: Yes, and I'll enrich the description of `expiration`.

cwen0 (approved):
LGTM

PR #189, opened 2022-07-27
Use shirou/gopsutils to get process full name

### LINKED ISSUE #186: shrink the name of process
The `chaosd` compared the specified process name with `comm`. However, the `comm` is limited to 16 length. If you have a process with longer name (e.g. "telegram-desktop"), it will be cutted into "telegram-deskto", which is unexpected by the users.

We could cut the process name arguments for the users, and compare the cutted argument with `comm`. However, it may cause unexpected affect (killing process with other name).

FingerLeader:

Close #186

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/joomcode/errorx v1.0.1
 	github.com/magiconair/properties v1.8.5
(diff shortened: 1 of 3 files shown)


YangKeao (approved):
LGTM

The implementations are nearly the same.

Andrewmatilde (approved):
LGTM

PR #198, opened 2022-08-04
download memStress and stress-ng according to the architecture

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

Close https://github.com/chaos-mesh/chaosd/issues/196

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,24 @@ endif
 PACKAGE_LIST := go list ./... | grep -vE "chaos-daemon/test|pkg/ptrace|zz_generated|vendor"
 PACKAGE_DIRECTORIES := $(PACKAGE_LIST) | sed 's|github.com/chaos-mesh/chaosd/||'
 
+UNAME_M := $(shell uname -m)
+ifeq ($(UNAME_M),x86_64)
+    ARCH_STRESS_NG = amd64
+	ARCH_MEM_STRESS = x86_64
+endif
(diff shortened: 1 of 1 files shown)


FingerLeader:
Should I change the target arch of `memStress` in the feature? So that both `stress-ng` and `memStress` can use the same tag.

YangKeao:
> Should I change the target arch of `memStress` in the feature? So that both `stress-ng` and `memStress` can use the same tag.

Good!

FingerLeader (approved):
LGTM

Andrewmatilde (approved):
LGTM

FingerLeader:
> > Should I change the target arch of `memStress` in the feature? So that both `stress-ng` and `memStress` can use the same tag.
> 
> Good! But you'll also need to modify the chaos mesh.
> 
> Or should I modify the target arch of `stress-ng`? (though they are the same with arch in matrix.)

Considering chaos mesh, maybe change the arch tag of `stress-ng` is better, cause `memStress` has the tag with other tools.

STRRL:
HI @YangKeao @FingerLeader @Andrewmatilde, could we merge this PR now?

YangKeao:
> HI @YangKeao @FingerLeader @Andrewmatilde, could we merge this PR now?

I'll change the architecture tag of the `stress-ng`, so don't merge.

PR #215, opened 2022-09-13
Store iptables when creating network attack

### LINKED ISSUE #213: network partition doesn't work when network delay was applied in advance

FingerLeader:
Signed-off-by: root <root@localhost.localdomain>
Close #213

### DIFF
--- a/pkg/server/chaosd/network.go
+++ b/pkg/server/chaosd/network.go
@@ -74,11 +74,10 @@ func (networkAttack) Attack(options core.AttackConfig, env Environment) (err err
 			}
 		}
 
-		if attack.NeedApplyIptables() {
-			if err = env.Chaos.applyIptables(attack, ipsetName, env.AttackUid); err != nil {
-				return perrors.WithStack(err)
-			}
+		if err = env.Chaos.applyIptables(attack, ipsetName, env.AttackUid); err != nil {
+			return perrors.WithStack(err)
 		}
+
 		// Because some tcs add filter iptables which will not be stored in the DB, we must re-apply these tcs to add the iptables.
 		if err = env.Chaos.applyTC(attack, ipsetName, env.AttackUid); err != nil {
 			return perrors.WithStack(err)
@@ -389,10 +388,6 @@ func (networkAttack) Recover(exp core.Experiment, env Environment) error {
 	case core.NetworkPortOccupiedAction:
 		return env.Chaos.recoverPortOccupied(attack, env.AttackUid)
 	case core.NetworkDelayAction, core.NetworkLossAction, core.NetworkCorruptAction, core.NetworkDuplicateAction, core.NetworkPartitionAction, core.NetworkBandwidthAction:
-		// `chaosdaemon.DeamonServer.SetTcs()` may build new iptables which will not be recorded in DB,
-		// and network partition is not suppose to build iptables directly, `recoverIptables()` will not
-		// be called when recovering a partition experiment. To avoid other cross-build situations, all these
-		// three functions will be called.
 		if err := env.Chaos.recoverIPSet(env.AttackUid); err != nil {
 			return perrors.WithStack(err)
 		}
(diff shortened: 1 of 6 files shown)


cwen0:
@FingerLeader I test this pr on my environment, I found the network delay still cancel when I applied a partition action.  

```
./bin/chaosd version
Chaosd Version: version.Info{GitVersion:"fix-partition-g83eccfacc55fe7", GitCommit:"83eccfacc55fe7a29c102df7cb05daa2ea1ac00f", BuildDate:"2022-09-13T15:03:27Z", GoVersion:"go1.19", Compiler:"gc", Platform:"linux/amd64"} 
./bin/chaosd attack network delay -i 39.156.66.10 -d eth0 -l 200ms
./bin/chaosd attack network partition -i 140.205.220.96 -d eth0
```
image

FingerLeader:
> I found the network delay still cancel when I applied a partition action.
> 
> ```
> ./bin/chaosd version
> ```

I test it again and find it works. Maybe you can delete `./bin/chaosd.dat` and try again.

cwen0:
> > I found the network delay still cancel when I applied a partition action.
> > ```
> > ./bin/chaosd version
> > ```
> 
> I test it again and find it works. Maybe you can delete `./bin/chaosd.dat` and try again.

OK, I will test it again

pkg/server/chaosd/network.go
```diff
@@ -79,11 +79,9 @@ func (networkAttack) Attack(options core.AttackConfig, env Environment) (err err
 				return perrors.WithStack(err)
```
cwen0, line 79: We can delete `NeedApplyIptables()` and `NeedApplyIPSet` directly.

pkg/server/chaosd/network.go
```diff
@@ -380,22 +389,20 @@ func (networkAttack) Recover(exp core.Experiment, env Environment) error {
 	case core.NetworkPortOccupiedAction:
 		return env.Chaos.recoverPortOccupied(attack, env.AttackUid)
 	case core.NetworkDelayAction, core.NetworkLossAction, core.NetworkCorruptAction, core.NetworkDuplicateAction, core.NetworkPartitionAction, core.NetworkBandwidthAction:
-		if attack.NeedApplyIPSet() {
-			if err := env.Chaos.recoverIPSet(env.AttackUid); err != nil {
-				return perrors.WithStack(err)
-			}
+		// `chaosdaemon.DeamonServer.SetTcs()` may build new iptables which will not be recorded in DB,
```
cwen0, line 392: We should update this comment.

pkg/server/chaosd/network.go
FingerLeader: I think leaving them can make the code logic clearly, and they may be needed to add features later (by the way, `NeedApplyIPSet()` is working now)

pkg/server/chaosd/network.go
FingerLeader: I don't think this comment should be updated, the filter iptables still not be stored in the DB

pkg/server/chaosd/network.go
cwen0: I think the logic will be clearer If we delete them. The logic: Chaos Mesh will reset all network action when you apply the network actions. This logic will be more simple.

pkg/server/chaosd/network.go
cwen0: Yes, but here we recover all because chaos mesh will apply all network actions when we apply a new one.

pkg/server/chaosd/network.go
FingerLeader: OK, I think it's OK to be the same as chaos mesh, `NeedApplyIptables` can be deleted, but I find `chaosd attack network delay -l 1s -d eth0` will not work without `NeedApplyIPSet()`

pkg/server/chaosd/network.go
FingerLeader: I agree

cwen0 (approved):
LGTM

PR #252, opened 2023-11-22
fix: device name always eth0 in iptable network attack partition

yuriscott:
In some scenarios, eth0 does not exist, but eno1 set through network attack partition will been replaced eth0 in Iptables.

### DIFF
--- a/pkg/core/network.go
+++ b/pkg/core/network.go
@@ -520,20 +520,20 @@ func (n *NetworkCommand) NeedApplyTC() bool {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


g1eny0ung (approved):
Hi @yuriscott, sorry for the late reply. I saw this PR in the Slack channel because someone mentioned it. It's LGTM.

But I would like to know if you have tested this PR in a real environment? like servers or local development environments. I will merge this PR within 1 day as I want to test it as well.

